PR #3, opened 2024-05-31
ci: update CI tooling

astapleton:
Use rustup and cargo directly instead of deprecated actions-rs actions. Also fix up the blinky example and remove unnecessary `power` module under `/examples` to get CI tests to pass.


.github/workflows/ci.yml
```diff
@@ -18,35 +18,30 @@ jobs:
           - stm32h562
           - stm32h563
           - stm32h573
+        logging:                # Example logging options
```
richardeoin, line 21: This is now testing all permutations of {rust, mcu, logging}, comment on line 12 needs to be updated. This might be excessive, I found that exercising the logging options for only one rust version is sufficient. But also ok to leave

richardeoin (approved):
Looks good to me, one small comment above

.github/workflows/ci.yml
astapleton: Yep. this is definitely overkill. It's also doing the same thing on nightly. I looked at the H7 CI and it looks like it's only testing the logging variations on the nightly and against just the one MCU. That seems reasonable. I'll do the same here.

PR #4, opened 2024-05-31
deps: Update dependencies

astapleton:
Update dependencies. Notably:
- update embedded-hal to 1.0.0, which is what will be developed against
- use version of stm32h5 available on crates.io (0.15.1)
- Add `defmt` as an optional dependency, to be used later
- Add `paste`, to be used in subsequent PRs


richardeoin:
Looks good to me!

richardeoin (approved):
Looks good to me!

PR #7, opened 2024-06-07
pwr: Add power configuration module

astapleton:
The Power configuration module manages voltage scaling modes, which need to be set correctly based on power usage considerations, as well as desired clock speed. The blinky example incorporates the power module and demonstrates choosing a voltage scaling mode.


src/pwr.rs
```diff
@@ -0,0 +1,169 @@
...
+//!
+//! After reset the system is in VOS3. This power configuration module allows
+//! modes VOS0 to VOS3 to be selected (it defaults to VOS0 after calling
+//! `freeze`.
+//!
+//! ```rust
+//!     let dp = pac::Peripherals::take().unwrap();
+//!
+//!     let pwr = dp.PWR.constrain();
+//!     let pwrcfg = pwr.freeze();
```
richardeoin, line 16: I would include a builder method in the example. Then in the documentation can state something like "if no mode is specified, then the default after calling `freeze` is VOS0"

```suggestion
//!     let pwrcfg = pwr.vos0().freeze();
```

src/pwr.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+/// Internal power methods
+impl Pwr {
+    /// Transition between voltage scaling levels
+    fn voltage_scaling_transition(&self, new_scale: VoltageScale) {
+        // ************************************
+        // TODO: Warning! STM32H503 Errata 2.2.13 states that for Rev A, Z this transition should only
+        //       be executed from RAM
+        //       Move this function into RAM!
```
richardeoin, line 107: This is a pretty impressive errata! I wonder if any Rev A or Z hardware ever made it out into the wild? Assuming that current dev boards have revision Y parts on them, I wouldn't worry about it too much.

richardeoin (approved):

src/pwr.rs
astapleton: Yeah, it's a fun one. I don't know whether Rev A or Z was ever released widely. It seems as though the dev boards that ST sells and the processors that I've used since Q3 2023 are Rev Y. I can probably just leave a note here, instead of a TODO, given how unlikely it seems that someone would actually run into this.

src/pwr.rs
astapleton: Good idea. I'll do so.

PR #9, opened 2024-06-07
rcc: Add reset reason calculation determination

astapleton:
The reset_reason submodule is used to determine the reset reason after a reset. 

As with the rest of the RCC code, this is adapted from same module in stm32h7xx-hal.


src/rcc/reset_reason.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    #[cfg(feature = "rm0492")]
+    // See R0492 Section 10.3.4 Reset source identification
+    return match (
+        reset_reason.lpwrrstf().is_reset_occurred(),
+        reset_reason.wwdgrstf().is_reset_occurred(),
+        reset_reason.iwdgrstf().is_reset_occurred(),
+        reset_reason.sftrstf().is_reset_occurred(),
+        reset_reason.pinrstf().is_reset_occurred(),
+        reset_reason.borrstf().is_reset_occurred(),
```
richardeoin, line 23: I think `pinrstf` and `borrstf` are swapped here. If swapped then it would match the order of Table 64

src/rcc/reset_reason.rs
```diff
@@ -0,0 +1,139 @@
...
+        (false, false, false, true, false, true) => {
+            ResetReason::SystemReset
+        }
+        (false, true, false, false, false, false) | (false, true, false, false, false, true) => {
+            ResetReason::WindowWatchdogReset
+        }
+        (false, false, true, false, false, false) => {
+            ResetReason::IndependentWatchdogReset
+        }
+        (true, false, false, false, false, false) => {
```
richardeoin, line 40: These don't quite agree with the table in section 10.3.4. They won't match if PINRSTF is set. Is this deliberate?

astapleton: Not deliberate. I think there are a few copy-pasta errors from the H7 codebase in this table.

astapleton:
@richardeoin Do the fixes/changes I pushed looked good?

richardeoin:
Great, this matches the RM!

richardeoin (approved):

PR #12, opened 2024-06-07
rcc: Add peripheral reset and control

astapleton:
Adds helper functions for enabling, resetting and disabling peripherals via the RCC control interface.

Note: This also disables builds for the STM32H56/7x processors as the STM32H5 PAC crate does not provide the necessary definitions for the RCC at this time.

### DIFF
--- a/src/rcc/rec.rs
+++ b/src/rcc/rec.rs
@@ -1,0 +1,551 @@
+//! Peripheral Reset and Enable Control (REC)
+//!
+//! This module contains safe accessors to the RCC functionality for each
+//! peripheral.
+//!
+//! At a minimum each peripheral implements
+//! [ResetEnable](trait.ResetEnable.html). Peripherals that have an
+//! individual clock multiplexer in the PKSU also have methods
+//! `kernel_clk_mux` and `get_kernel_clk_mux`. These set and get the state
+//! of the kernel clock multiplexer respectively.
+//!
+//! Peripherals that share a clock multiplexer in the PKSU with other
+//! peripherals implement a trait with a `get_kernel_clk_mux` method that
+//! returns the current kernel clock state. Because the kernel_clk_mux is shared
+//! between multiple peripherals, it cannot be set by any individual one of
+//! them. Instead it can only be set by methods on the
+//! [`PeripheralRec`](struct.PeripheralREC.html) itself. These methods are named
+//! `kernel_xxxx_clk_mux()`.
+//!
+//! # Reset/Enable Example
+//!
+//! ```
+//! // Constrain and Freeze power
+//! ...
+//! let rcc = dp.RCC.constrain();
+//! let ccdr = rcc.sys_ck(100.MHz()).freeze(pwrcfg, &dp.SYSCFG);
+//!
+//! // Enable the clock to a peripheral and reset it
+//! ccdr.peripheral.FDCAN.enable().reset();
+//! ```
+//!
+//! # Individual Kernel Clock Example
+//! ```
+//! let ccdr = ...; // Returned by `freeze()`, see example above
+//!
+//! // Set individual kernel clock
+//! let cec_prec = ccdr.peripheral.CEC.kernel_clk_mux(CecClkSel::LSI);
+//!
+//! assert_eq!(cec_prec.get_kernel_clk_mux(), CecClkSel::LSI);
+//! ```
+//!
+//! # Group Kernel Clock Example
+//! ```
+//! let mut ccdr = ...; // Returned by `freeze()`, see example above
+//!
+//! // Set group kernel clock mux
+//! ccdr.peripheral.kernel_i2c123_clk_mux(I2c123ClkSel::PLL3_R);
+//!
+//! // Enable and reset peripheral
+//! let i2c3_prec = ccdr.peripheral.I2C3.enable().reset();
+//!
+//! assert_eq!(i2c3_prec.get_kernel_clk_mux(), I2c123ClkSel::PLL3_R);
+//!
+//! // Some method that consumes the i2c3 prec
+//! init_i2c3(..., i2c3_prec);
+//!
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -15,9 +15,9 @@ jobs:
           - stable
         mcu:
           - stm32h503
-          - stm32h562
-          - stm32h563
-          - stm32h573
+          # - stm32h562
+          # - stm32h563
+          # - stm32h573
```
astapleton, line 20: This was disappointing to disable so soon... I had forgotten that when I did the RCC in stm32-rs last year, I only focussed on the STM32H503 RCC module because of how much there is to it, and how different the STM32H562/63/73 were. I'll have to get back to that to be able to get this code working on the rest of the family.

joaomorenorf:
Hello, how can I help in this?
I have a special interest in the STM32H573.

astapleton:
> Hello, how can I help in this? I have a special interest in the STM32H573.

Hey @joaomorenorf! So I based this on top of the definitions I added to the [stm32-rs](https://github.com/stm32-rs/stm32-rs) repo to flesh out the peripheral definitions for the H503 (I mostly tried to include definitions for the H573 and H562/3 where they didn't significantly diverge from from the H503, but unfortunately the RCC was one case where it did). 

You could make a PR into  repo to add the definitions for the RCC for the STM32H573 (and STM32H562/3), as they share the same RCC definitions). 

You can see what I've done with H503 here:
- H503 device: https://github.com/stm32-rs/stm32-rs/blob/master/devices/stm32h503.yaml#L31-L32
- I split the RCC register naming fixes into a common H5 and a H503 specific file:
  - https://github.com/stm32-rs/stm32-rs/blob/master/devices/common_patches/rcc/h5.yaml
  - https://github.com/stm32-rs/stm32-rs/blob/master/devices/common_patches/rcc/h503.yaml
- And did the same with the field definitions:
  - https://github.com/stm32-rs/stm32-rs/blob/master/peripherals/rcc/rcc_h5.yaml
  - https://github.com/stm32-rs/stm32-rs/blob/master/peripherals/rcc/rcc_h503.yaml

As I noted in a few of those files, the H573 is mostly a superset of the H503 in terms of functionality, so it should just be necessary to add the missing register definitions.

That all said... it took a while for my PRs to get merged into the `stm32-rs` mainline and published. While I think that is the better long term approach, and I would recommend you do so anyway, it would probably be possible to use the RCC definitions as provided by the SVD (ie. unpatched by stm32-rs) as a stopgap. It might be possible to adapt what I have here to use those definitions, or we could include a separate module that does the same thing for the H573 and H562/3 until they can be unified. 

I can take a look at whether I can adapt what I have here easily, but I won't likely be able to get to it for the next week or so. Otherwise, if you wanted to take a stab at one or the other approach, that would be awesome, and I would be very happy to give you pointers, review etc.

astapleton:
@joaomorenorf So I took another look at what was breaking the build, and realized that the main breakage was just that the clock source variants that let you easily select which clock to use for which peripheral are only defined for the H503. I reworked the macro in `rec` and was able to move those all into guards specific to the H503, and compilation now works for all variants!

There's one caveat: this means that to use one of these peripherals there would be the minor inconvenience of using the PAC directly to configure the clock source when a non-default clock is used for a peripheral, and when instantiating a peripheral driver such as I2C, you'd need to provide a dummy implementation of the ResetEnable trait (this will become more apparent as I merge these drivers in, but it's a pretty minor inconvenience, IMO) and everything should be otherwise quite useable.

I would still recommend making a PR into the `stm32-rs` repo (not sure when I'd get to it, tbh), and we can then incorporate the H573 peripherals into this module.

astapleton:
@joaomorenorf On another note, I'll try to make sure that the code compiles for all variants (and keep them enabled in CI!), but I haven't _tested_ anything on those variants, and there is a bunch of functionality that I haven't implemented (both for other MCUs in the family and in general). As an example in the RCC module, PLL3 looks like it has a subset of the functionality of the other 2 PLLs, so we'd need to account for that before we can enable the PLL. 

As for how you could help further, testing the code on those MCUs would be super helpful, and your contributions to this repo for missing functionality would be very welcome too. I've still got a bunch of modules still to merge (e.g. serial, gpio, spi, i2c, icache), but there's a lot more that needs to be implemented, so there's plenty of scope to help.

joaomorenorf:
Thank you, @astapleton, Now I'll focus on adding the registers definitions to `stm32-rs`.

src/rcc/rec.rs
```diff
@@ -0,0 +1,551 @@
...
+
+    #[cfg(all())]
+    APB3, "Advanced Peripheral Bus 3 (APB3) peripherals" => [
+        Sbs
+    ];
+    #[cfg(feature = "rm0492")]
+    APB3, "" => [
+        I3c2 [kernel clk: I3c2(Variant) ccipr4 "I3C2"],
+        LpTim1 [kernel clk: LpTim1(Variant) ccipr2 "LPTIM1"],
+        LpUart1 [kernel clk: LpUart1(Variant) ccipr3 "LPUART1"]
```
richardeoin, line 548: There's also the `RTCAPBEN` bit here. On H7 this was enabled by default after reset, so the HAL ignores it. If I understand the RM correctly, on H5 it will need to be enabled to access the RTC registers.

richardeoin (approved):

src/rcc/rec.rs
astapleton: Looks like I excluded this, intentionally or not, because the RTC isn't resettable via the RCC. I'm looking for a way to account for that in the macro definition.

astapleton: I updated the macro to handle the cases where the peripheral is not resettable, and added missing peripherals, including the RTC

richardeoin: Great that those methods are available now!

richardeoin (approved):

PR #13, opened 2024-06-07
rcc: Calculate PLL clocks and implement core clock configuration logic

astapleton:
This implements the PLL and clock configuration logic for the processor using the RCC peripheral. 
- The `pll` module is responsible for configuring the VCOs and setting up the PLL dividers for the desired frequency
- The `freeze` method in the `rcc` module does the heavy lifting for making sure that the processor clocks are configured in the desired manner

Note: This is largely copied from the implementation in stm32h7xx-hal, although it significantly refactors and modifies the PLL configuration in the `pll` module. This was mostly due to the differences in the register definitions used by the calculation and the fact that the H5 family (and especially the H503 MCU) is quite a bit simpler than the H7 family. I made significant changes to the the PLL configuration due to the stm32h7xx-hal crate being a bit bugged: it calculates values for Q and R outputs that are invalid for fractional PLL configurations.

### DIFF
--- a/src/rcc/pll.rs
+++ b/src/rcc/pll.rs
@@ -1,0 +1,572 @@
+//! Phase Locked Loop Configuration
+
+use core::ops::RangeInclusive;
+
+use super::{Rcc, HSI};
+use crate::stm32::rcc::pll1cfgr::PLL1VCOSEL as VCOSEL;
+use crate::stm32::RCC;
+use crate::time::Hertz;
+
+const FRACN_DIVISOR: f32 = 8192.0; // 2 ** 13
+const FRACN_MAX: f32 = FRACN_DIVISOR - 1.0;
+
+const PLL_N_MIN: u32 = 4;
+const PLL_N_MAX: u32 = 512;
+const PLL_OUT_DIV_MAX: u32 = 128;
+const PLL_M_MAX: u32 = 63;
+
+/// Strategies for configuring a Phase Locked Loop (PLL)
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub enum PllConfigStrategy {
+    /// No fractional component
+    Integer,
+    /// PLL configured with fractional divider
+    Fractional,
+}
+
+/// Configuration of a Phase Locked Loop (PLL)
+pub struct PllConfig {
+    pub(super) strategy: PllConfigStrategy,
+    pub(super) p_ck: Option<u32>,
+    pub(super) q_ck: Option<u32>,
+    pub(super) r_ck: Option<u32>,
+}
+impl Default for PllConfig {
+    fn default() -> PllConfig {
+        PllConfig {
+            strategy: PllConfigStrategy::Integer,
+            p_ck: None,
+            q_ck: None,
+            r_ck: None,
+        }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


src/rcc/pll.rs
```diff
@@ -0,0 +1,562 @@
...
+        let error = output_p - pll_p_target as f32;
+        assert!(
+            f32::abs(error) < (pll_p_target as f32 / 500_000.0),
+            "P error too large: {error}"
+        ); // < ±.0002% = 2ppm error
+
+        let output_q = vco_ck_achieved as f32 / pll_x_q as f32;
+        println!("Q Divider {}", pll_x_q);
+        println!("==> Output Q {} MHz", output_q / 1e6);
+        println!();
```
richardeoin, line 555: Ideally there would be an `assert!` statements to check the values here

richardeoin:
This is by necessity quite complex and error-prone logic, and at a first review it looks good! The two options for `PllConfigStrategy` cover most use cases and keeps in simple. Can you explain a little more about the bug you saw with Q and R outputs for fractional dividers? The Q and R divider values are calculated from the vco output frequency, which should be valid in either integer or fractional mode

richardeoin:
This is by necessity quite complex and error-prone logic, and at a first review it looks good! The two options for `PllConfigStrategy` cover most use cases and keeps it simple. Can you explain a little more about the bug you saw with Q and R outputs for fractional dividers? The Q and R divider values are calculated from the vco output frequency, which should be valid in either integer or fractional mode

astapleton:
> This is by necessity quite complex and error-prone logic, and at a first review it looks good! The two options for `PllConfigStrategy` cover most use cases and keeps it simple. Can you explain a little more about the bug you saw with Q and R outputs for fractional dividers? The Q and R divider values are calculated from the vco output frequency, which should be valid in either integer or fractional mode

Looking back at the patch I originally made to address the error I found, it looks like because the H7 VCO derivation chooses the maximum output frequency for the VCO (see [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/rcc/pll.rs#L130-L133)), it looks like if a high enough frequency is chosen for the system clock, which is then derived from PLL1, it's not possible to select frequencies in the low megahertz range on a different PLL (divider is >128). It looks like the patch I made (last year, so my memory is a bit fuzzy) ended up becoming quite a bit more extensive than just addressing that, so the VCO setup here is quite a bit different than the H7 code. For instance, the `fractional-pll` example in #15 would not work with the H7 logic. It might be that I introduced an error while adapting it for H5, so perhaps it does work fine on an actual H7.

richardeoin:
> the H7 VCO derivation chooses the maximum output frequency for the VCO (see [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/rcc/pll.rs#L130-L133)), it looks like if a high enough frequency is chosen for the system clock, which is then derived from PLL1, it's not possible to select frequencies in the low megahertz range on a different PLL (divider is >128).

Aha, I ignored the use case of PLL outputs with frequencies of the (max VCO frequency/128) or less. Nice that you've improved that.

Thanks for adding the tests.

richardeoin (approved):

PR #17, opened 2024-07-31
gpio: Add basic functionality for GPIOs

astapleton:
This adds the basic outline functionality for the GPIO module and will form the base for follow-up PRs (see #19, #20, #21)  to round out the GPIO functionality. 

The gpio module is largely based on/copied from the implementation in [stm32h7xx-hal](https://github.com/stm32-rs/stm32h7xx-hal).

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -1,0 +1,447 @@
+//! General Purpose Input / Output
+//!
+//! The GPIO pins are organised into groups of 16 pins which can be accessed through the
+//! `gpioa`, `gpiob`... modules. To get access to the pins, you first need to convert them into a
+//! HAL designed struct from the `pac` struct using the [split](trait.GpioExt.html#tymethod.split) function.
+//! ```rust
+//! // Acquire the GPIOA peripheral
+//! // NOTE: `dp` is the device peripherals from the `PAC` crate
+//! let mut gpioa = dp.GPIOA.split();
+//! ```
+//!
+//! This gives you a struct containing all the pins `px0..px15`.
+//! By default pins are in floating input mode. You can change their modes.
+//! For example, to set `pa5` high, you would call
+//!
+//! ```rust
+//! let output = gpioa.pa5.into_push_pull_output();
+//! output.set_high();
+//! ```
+//!
+//! ## Modes
+//!
+//! Each GPIO pin can be set to various modes:
+//!
+//! - **Alternate**: Pin mode required when the pin is driven by other peripherals
+//! - **Analog**: Analog input to be used with ADC.
+//! - **Dynamic**: Pin mode is selected at runtime. See changing configurations for more details
+//! - Input
+//!     - **PullUp**: Input connected to high with a weak pull up resistor. Will be high when nothing
+//!       is connected
+//!     - **PullDown**: Input connected to high with a weak pull up resistor. Will be low when nothing
+//!       is connected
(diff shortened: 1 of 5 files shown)


src/gpio.rs
```diff
@@ -0,0 +1,447 @@
...
+pub enum Edge {
+    /// Rising edge of voltage
+    Rising,
+    /// Falling edge of voltage
+    Falling,
+    /// Rising and falling edge of voltage
+    RisingFalling,
+}
+
+pub type AF0<Otype = PushPull> = Alternate<0, Otype>;
```
richardeoin, line 211: Nice tidy-up of the macro that was not saving any lines, although the docstrings are lost.

The docstrings looked ok in the docs, but not a big issue
image

richardeoin:
Great that this GPIO implementation is applicable to H5 also! The H7 implementation this is based on was mostly contributed by @burrbull in https://github.com/stm32-rs/stm32h7xx-hal/pull/347 and https://github.com/stm32-rs/stm32h7xx-hal/pull/346

I haven't had time to do an in-depth review (limited time at the moment) but what I have seen is good :)

burrbull:
@astapleton Good for initial stuff.

You could look at https://github.com/stm32-rs/stm32f4xx-hal/tree/master/src/gpio if you want to develop it further.

astapleton:
> @astapleton Good for initial stuff.
> 
> You could look at https://github.com/stm32-rs/stm32f4xx-hal/tree/master/src/gpio if you want to develop it further.

Wow, thanks for the quick response! Would you mind if I added you as reviewer to the other GPIO PRs I have? They include some of the other functionality you mentioned. I haven't included some of the smarter GPIO pin allocation stuff that the M4 library does because it seems like a big change that would have impacts on the downstream work that I've already done and want to push upstream before I start making additional changes like that.

burrbull:
> It seems that quite a few of the peripherals are very similar between the H7 and H5 (with some notable differences like a completely new DMA engine on the H5).

Was it really needed to create separate repo/crate from stm32h7xx-hal?
Maybe just add H5 support there?

astapleton:
> > It seems that quite a few of the peripherals are very similar between the H7 and H5 (with some notable differences like a completely new DMA engine on the H5).
> 
> Was it really needed to create separate repo/crate from stm32h7xx-hal? Maybe just add H5 support there?

While a number of the peripherals are quite similar, some are also quite different: the H5 has a completely new DMA engine (GPDMA). In other ways the families are also quite different. The H5 family is all Cortex-M33 single core MCUs, vs the very complex multi-core H7s with mixed Cortex-M7/M4 cores. It introduces a slew of new security peripherals/modules, which can gate access to peripherals and other functionality, as well as support for some sort of Trusted Execution Environment implementation on a subset of the cores. It also uses a completely new system control block (SBS). 

All that said, I think there's enough difference between the families to justify a different HAL to avoid adding significant complexity to the stm32h7 crate. I think if I was looking for a way of sharing code between the various ST modules, I think moving more of the peripheral control into separate crates would probably be my preferred solution. The SPI/I2C/GPIO/UART peripherals are quite similar across a number of families, not just the H7 and H5 families.

src/gpio.rs
astapleton: Good catch on the docstrings. I added them back. It's a little less neat, but I'm still not convinced that a macro is necessary.

PR #20, opened 2024-08-13
gpio: add dynamic pin support and conversion functions

astapleton:
- Add DynamicPin to allow for dynamic pin mode changes
- Add type conversion functions for pin configuration

This is taken directly from the implementation in [stm32h7xx-hal](https://github.com/stm32-rs/stm32h7xx-hal).

### DIFF
--- a/src/gpio/convert.rs
+++ b/src/gpio/convert.rs
@@ -126,22 +126,6 @@ where
     }
 }
 
-macro_rules! af {
-    ($($into_alternate_af:ident: $A:literal;)+) => {
-        $(
-            #[doc="Configures the pin to operate in alternate function "]
-            #[doc=stringify!($A)]
-            #[doc=" mode"]
-            #[deprecated(since = "0.12.0", note = "Use the .into_alternate() method instead")]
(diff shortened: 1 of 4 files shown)


src/gpio/convert.rs
```diff
@@ -0,0 +1,477 @@
...
+    }
+}
+
+macro_rules! af {
+    ($($into_alternate_af:ident: $A:literal;)+) => {
+        $(
+            #[doc="Configures the pin to operate in alternate function "]
+            #[doc=stringify!($A)]
+            #[doc=" mode"]
+            #[deprecated(since = "0.12.0", note = "Use the .into_alternate() method instead")]
```
ryan-summers, line 135: I don't think there should be deprecated functions introduced here ;)

ryan-summers (changes requested):

src/gpio/convert.rs
astapleton: Oops. Copy paste error. Thanks for catching!

astapleton:
The version check failing is addressed by #25

ryan-summers (approved):

PR #22, opened 2024-08-22
icache: Add ICACHE management

astapleton:
Add management functions for the ICACHE instruction cache peripheral.

### DIFF
--- a/src/icache.rs
+++ b/src/icache.rs
@@ -1,0 +1,200 @@
+//! Instruction Cache management
+//!
+//! To enable the the ICACHE:
+//!
+//! ```
+//! let dp = ...;            // Device peripherals
+//! let mut icache = dp.ICACHE.constrain();
+//!
+//! icache.enable();
+//! ```
+//!
+//! To invalidate the cache, use the ICache::invalidate() method:
+//!
+//! ```
+//! icache.invalidate();
+//! ```
+//!
+//! Performance monitoring of the cache is possible using the cache hit and
+//! miss counters:
+//!
+//! ```
+//! icache.enable_hit_counter();
+//! icache.enable_miss_counter();
+//!
+//! // do something
+//!
+//! let hit_count = icache.hit_count();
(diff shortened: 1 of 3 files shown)


src/icache.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    /// Enable ICACHE in default operating mode (N-way associative)
+    pub fn enable(&mut self) {
+        self.enable_n_way()
+    }
+
+    /// Enable the ICACHE in N-way associative mode
+    pub fn enable_n_way(&mut self) {
+        if self.is_enabled() {
+            return;
```
ryan-summers, line 71: Shouldn't we return an error so the user knows to do something different instead of silently ignoring their request? Same for all the modes below. Even returning `Err(())` might be enough.

astapleton: Good point about ignoring the request. I'll have a look at the peripheral again (I wrote this a while ago) and see what would be appropriate. I'm thinking that maybe switching gracefully between N-way and direct mapped modes might be better for this and the `enable_direct_mapped` mode. I think for `disable` simply returning if it is already disabled seems fine: it might be called more than once, and that's fine I think?

For the reset functions on the counters, I think the counter is reset if the peripheral is disabled so that's probably fine, but I'll double check. The asserts on enabling the hit counter are more aggressive than other calls where things are ignored, so I can take a look at those.

astapleton:
@ryan-summers Ok, I've reworked this API significantly. It removes a lot of the edge cases for behavior when the ICACHE is disabled vs enabled or what mode it is in by having the `ICache` struct only be valid while the peripheral is enabled. Enabling it in either mode gives you the `ICache` struct, and disabling the peripheral drops it returning the peripheral struct `ICACHE` that can then be re-enabled. I also added interrupt enable/disable and servicing methods.

ryan-summers (approved):
That looks much nicer from a design perspective! :)

PR #23, opened 2024-10-04
delay: Add delay module

astapleton:
This implements the embedded-hal 1.0 `DelayNs` trait. The cortex-m crate has essentially the same logic, but the published version is based on embedded-hal 0.2. This can be removed (repurposed for use with timers?) when a new version of the cortex-m crate is published.

The blinky example has been updated to use the delay module.

### DIFF
--- a/examples/blinky.rs
+++ b/examples/blinky.rs
@@ -4,12 +4,15 @@
 mod utilities;
 
 use cortex_m_rt::entry;
-use stm32h5xx_hal::{pac, prelude::*};
+use embedded_hal::delay::DelayNs;
+use fugit::SecsDurationU32;
+use stm32h5xx_hal::{delay::Delay, pac, prelude::*, rcc::ResetEnable};
 
 #[entry]
 fn main() -> ! {
     utilities::logger::init();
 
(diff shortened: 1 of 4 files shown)


ryan-summers (approved):

ryan-summers:
Have you done any timing tests on the implementation to make sure it's correct?

astapleton:
> Have you done any timing tests on the implementation to make sure it's correct?

I believe so, but I'll play around with the blinky example and double check that.

astapleton:
> Have you done any timing tests on the implementation to make sure it's correct?

I ran some tests and it's a few milliseconds fast per second if I sleep repeatedly for a second. Not sure if that's worth chasing down though: it could just be core clock inaccuracy (due to internal reference).

ryan-summers:
Hmm, a delay should always delay for the minimum time specified. If it can delay less, that sounds problematic

astapleton:
> Hmm, a delay should always delay for the minimum time specified. If it can delay less, that sounds problematic

I'll look further into it then

ryan-summers (approved):
Sounds good to me :) Feel free to merge when you feel its ready! Thanks for diving into it

PR #27, opened 2024-11-15
i2c: Add I2C Controller implementation

astapleton:
This adds the I2C Controller driver implementation for the STM32H5. It implements a blocking, and non-blocking API, as well as the embedded-hal 1.0 traits.

It borrows a lot from the STM32H7 HAL (particularly the TIMINGR configuration, which is a bit inscrutable in the reference manual), but it borrows the `Instance` concept from the STM32F4 project to minimize the need for macros. 

It uses the updated terminology of Controller/Target from the latest version of the I2C spec (Rev. 7.0)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,6 @@ embedded-hal = "1.0.0"
 defmt = { version = "0.3.8", optional = true }
 paste = "1.0.15"
 log = { version = "0.4.20", optional = true}
-nb = "1.1.0"
 
 [dev-dependencies]
 log = { version = "0.4.20"}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,3 +88,7 @@ opt-level = "s" # optimize for binary size
 [[example]]
 name = "blinky"
 required-features = ["stm32h503"]
+
+[[example]]
+name = "i2c"
+required-features = ["stm32h503"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,7 @@ embedded-hal = "1.0.0"
 defmt = { version = "0.3.8", optional = true }
 paste = "1.0.15"
 log = { version = "0.4.20", optional = true}
+nb = "1.1.0"
 
 [dev-dependencies]
 log = { version = "0.4.20"}
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -437,15 +437,15 @@ impl<I2C: Instance> I2c<I2C> {
             calc_timing_params(i2c_ker_ck, freq);
         i2c.timingr().write(|w| {
             w.presc()
-                .bits(presc_reg)
+                .set(presc_reg)
                 .scll()
-                .bits(scll)
+                .set(scll)
                 .sclh()
-                .bits(sclh)
+                .set(sclh)
                 .sdadel()
-                .bits(sdadel)
+                .set(sdadel)
                 .scldel()
-                .bits(scldel)
+                .set(scldel)
         });
 
         // Enable the peripheral and analog filter
@@ -534,7 +534,7 @@ impl<I2C: Instance> Inner<I2C> {
     fn flush_txdr(&mut self) {
         // If a pending TXIS flag is set, write dummy data to TXDR
         if self.i2c.isr().read().txis().bit_is_set() {
-            self.i2c.txdr().write(|w| w.txdata().bits(0));
+            self.i2c.txdr().write(|w| w.txdata().set(0));
         }
 
         // If TXDR is not flagged as empty, write 1 to flush it
@@ -582,7 +582,7 @@ impl<I2C: Instance> Inner<I2C> {
         self.check_clear_target_nack(&isr)?;
 
         if isr.txis().is_empty() {
-            self.i2c.txdr().write(|w| w.txdata().bits(data));
+            self.i2c.txdr().write(|w| w.txdata().set(data));
             Ok(())
         } else {
             Err(nb::Error::WouldBlock)
@@ -657,7 +657,7 @@ impl<I2C: Instance> I2c<I2C> {
             w.rd_wrn()
                 .bit(direction == Direction::Read)
                 .nbytes()
-                .bits(length as u8)
+                .set(length as u8)
                 .start()
                 .start()
                 .autoend()
@@ -667,9 +667,9 @@ impl<I2C: Instance> I2c<I2C> {
                 .add10()
                 .bit(address_mode == AddressMode::AddressMode10bit);
             if address_mode == AddressMode::AddressMode10bit {
-                w.sadd().bits(addr).head10r().complete();
+                w.sadd().set(addr).head10r().complete();
             } else {
-                w.sadd().bits(addr << 1);
+                w.sadd().set(addr << 1);
             }
             w
         });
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -50,6 +50,7 @@ embedded-hal = "1.0.0"
 defmt = { version = "0.3.8", optional = true }
 paste = "1.0.15"
 log = { version = "0.4.20", optional = true}
+nb = "1.1.0"
```
ryan-summers, line 53: I really would avoid using the `nb` crate nowadays. There's been talk of entirely deprecating it. It was intended to support embedded `async` before `async` was possible on embedded, but there's more modern approaches now with `async` frameworks targeting embedded.

src/i2c.rs
```diff
@@ -0,0 +1,1037 @@
...
+//!
+//! ```
+//! use nb::block;
+//!
+//! // Simple write
+//! let write = [0x11, 0x22, 0x33];
+//! i2c.start(0x18, AddressMode::AddressMode7bit, write.len(), Stop::Automatic)
+//!
+//! for byte in write {
+//!     block!(i2c.write_nb(byte))?;
```
ryan-summers, line 62: I'd remove the `nb` implementation and instead go for `embedded-hal-async` personally, but up to you

src/i2c.rs
```diff
@@ -0,0 +1,1037 @@
...
+        self.i2c.cr1().modify(|_, w| w.pe().enabled());
+    }
+
+    /// Start listening for interrupt `event`
+    pub fn listen(&mut self, event: Event) {
+        self.i2c.cr1().modify(|_, w| match event {
+            Event::Transmit => w.txie().enabled(),
+            Event::Receive => w.rxie().enabled(),
+            Event::Reload => w.tcie().enabled(),
+            Event::TransferComplete => w.tcie().enabled(),
```
ryan-summers, line 474: Are these (same as below) supposed to be looking at the same bit?

src/i2c.rs
```diff
@@ -0,0 +1,1037 @@
...
+            Event::Stop => w.stopie().enabled(),
+            Event::Error => w.errie().enabled(),
+            Event::NotAcknowledge => w.nackie().enabled(),
+            Event::AddressMatch => w.addrie().enabled(),
+        });
+        let _ = self.i2c.cr1().read();
+        let _ = self.i2c.cr1().read(); // Delay 2 peripheral clocks
```
ryan-summers, line 481: Would be helpful to document why there's a delay here

src/i2c.rs
```diff
@@ -0,0 +1,1037 @@
...
+        };
+        Ok(result)
+    }
+}
+
+impl<I2C: Instance> Inner<I2C> {
+    fn flush_txdr(&mut self) {
+        // If a pending TXIS flag is set, write dummy data to TXDR
+        if self.i2c.isr().read().txis().bit_is_set() {
+            self.i2c.txdr().write(|w| w.txdata().bits(0));
```
ryan-summers, line 537: Are you certain this won't enqueue data to be written over I2C?

src/i2c.rs
```diff
@@ -0,0 +1,1037 @@
...
+        }
+
+        // If TXDR is not flagged as empty, write 1 to flush it
+        if self.i2c.isr().read().txe().is_not_empty() {
+            self.i2c.isr().write(|w| w.txe().set_bit());
+        }
```
ryan-summers, line 543: Do you need to poll something until the flush completes?

ryan-summers (commented):
Have you tested these on the H5 as well? I'd be a bit suspicious around some of the timing calcs, so at least one functional test would give some confidence.

Main comments in this are around the use of nb. I didn't closely look into the register calcs or device datasheet

Cargo.toml
astapleton: Ah, thanks for this context. I'll look at `embedded-hal-async`

astapleton:
> Have you tested these on the H5 as well? I'd be a bit suspicious around some of the timing calcs, so at least one functional test would give some confidence.
> 
> Main comments in this are around the use of nb. I didn't closely look into the register calcs or device datasheet

The suspicion is warranted, I'll give you that. I honestly don't understand them properly and the datasheet doesn't explain them very well and just refers you to use STM32Cube to determine the timing values. However, I have tested these timing values quite extensively on the H503. I'm using them for a I2C target implementation (which I'll put up for review after I merge this) in pre-production hardware and they have worked well so far. I'll take another look a them though.

src/i2c.rs
astapleton: The Reload and TransferComplete both looking at TCIE? I'll double check

Cargo.toml
astapleton: @ryan-summers I looked at `embedded-hal-async`, but I'm not sure how it solves the same problem? It doesn't provide any tool to manage repeatedly trying a blocking call. Is the suggestion to remove non-blocking calls entirely? Wouldn't they be useful in interrupts, or for creating a Future implementation for async? 

Is it `embedded-hal-nb` that might be deprecated rather than the `nb` crate? That wasn't actually at 1.0 when I originally created this driver, so I didn't implement those APIs.

src/i2c.rs
astapleton: Page 1216 of RM0492 Rev 2 indicates that both of these events generate a TCIE interrupt.

src/i2c.rs
astapleton: It only writes to it after the transaction has finished (after a NACK), so no data will be written out. It's derived from stm32h7xx-hal, which added it here: https://github.com/stm32-rs/stm32h7xx-hal/pull/43. I've also added more calls to the `flush_txdr` method due to some controller mode edge cases I found, where unhandled state would cause the driver to lock up.

src/i2c.rs
astapleton: I believe what this does is clear the TXDR register such that when the next transaction is started, the TXIS bit gets set, which it wouldn't if the TXDR register was never flushed (and the transaction hangs waiting for TXIS).

astapleton:
I've looked at the TIMINGR calculations again, and I still don't know where some of those numbers come from, but between the tests, the fact that they've been working pretty well for me, and seem to do a good job for the H7 and the fact that embassy uses a very [similar derivation](https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/i2c/v2.rs#L709-L801) (probably copied from stm32h7xx-hal, based on timing and similarity), I think I'll stick with them as is.

astapleton:
I rebased this on top of master, and it looks like something broke with the new PAC and REC definitions. I'll take a look at that tomorrow.

Cargo.toml
astapleton: @ryan-summers Coming back to this, is the recommendation to declare non-blocking/async methods that return a `Future` instead of using the `nb` crate? Or have non-blocking methods that just check a status and return `bool`/ `Result<bool>` instead of returning an `nb::Error::WouldBlock` and then have methods that can't be easily decomposed to a bool return a `Future`? If you're not using an async runtime, how would one implement interrupts? Just manually invoking poll on the future?

ryan-summers: Generally, the way HALs do this is to have entirely blocking APIs for external `embedded-hal` traits. If you also want a non-blocking API, you'd use `embedded-hal-async` and could compose your software internally to handle both cases. `nb` doesn't really solve the problem at all.

astapleton:
@ryan-summers I've removed `nb` crate and the public non-blocking functions as you suggested, as well as anything to do with interrupts for which they would have been used. Should be ready to go now.

ryan-summers (approved):
LGTM! Nicely organized.

PR #30, opened 2025-01-07
rcc: add peripheral clock definitions for the rest of the H5 family

astapleton:
Add Peripheral clock + reset/enable definitions for the rest of the H5 family (for which we have SVDs - still waiting on the H523/H533 SVDs).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,11 +31,15 @@ rm0481 = ["gpio-h5x"]   # STM32H52x/6x/7x
 gpio-h503 = []
 gpio-h5x = []
 
+ethernet = []
+otfdec = []
+sdmmc2 = []
+
 rt = ["stm32h5/rt"]
 stm32h503 = ["stm32h5/stm32h503", "device-selected", "rm0492"]
 stm32h562 = ["stm32h5/stm32h562", "device-selected", "rm0481"]
-stm32h563 = ["stm32h5/stm32h563", "device-selected", "rm0481"]
-stm32h573 = ["stm32h5/stm32h573", "device-selected", "rm0481"]
+stm32h563 = ["stm32h5/stm32h563", "device-selected", "rm0481", "sdmmc2", "ethernet"]
+stm32h573 = ["stm32h5/stm32h573", "device-selected", "rm0481", "otfdec", "sdmmc2", "ethernet"]
 # Flags for examples
 log = ["dep:log"]
 log-itm = ["log"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,21 +25,28 @@ device-selected = []
 
 # rmXXXX represent processor subfamilies and their common features
 rm0492 = ["gpio-h503"]
-rm0481 = ["gpio-h5x"]   # STM32H52x/6x/7x
+rm0481 = []                 # STM32H52x/6x/7x
 
-# Different subfamilies have different GPIOs available
+# Different subfamily groups have different selections of identical peripherals
+h56x_h573 = ["gpio-h56x_h573"]         # STM32H56x/STM32H573
+h523_h533 = []                         # STM32H523/STM32H533
+
+# Different subfamily groups have different GPIOs available and different alternate functions
+# available
 gpio-h503 = []
-gpio-h5x = []
+gpio-h56x_h573 = []
 
-ethernet = []
-otfdec = []
-sdmmc2 = []
+# Some peripherals are only present on specific MCUs within a subfamily group
+ethernet = []               # Only STM32H563/73 have ethernet
+otfdec = []                 # Only STM32H573/33 have OTFDEC
+sdmmc2 = []                 # Only STM32H563/73 have SDMMC2
 
 rt = ["stm32h5/rt"]
 stm32h503 = ["stm32h5/stm32h503", "device-selected", "rm0492"]
-stm32h562 = ["stm32h5/stm32h562", "device-selected", "rm0481"]
-stm32h563 = ["stm32h5/stm32h563", "device-selected", "rm0481", "sdmmc2", "ethernet"]
-stm32h573 = ["stm32h5/stm32h573", "device-selected", "rm0481", "otfdec", "sdmmc2", "ethernet"]
+stm32h562 = ["stm32h5/stm32h562", "device-selected", "rm0481", "h56x_h573"]
+stm32h563 = ["stm32h5/stm32h563", "device-selected", "rm0481", "h56x_h573", "sdmmc2", "ethernet"]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -31,11 +31,15 @@ rm0481 = ["gpio-h5x"]   # STM32H52x/6x/7x
...
+otfdec = []
+sdmmc2 = []
+
 rt = ["stm32h5/rt"]
 stm32h503 = ["stm32h5/stm32h503", "device-selected", "rm0492"]
 stm32h562 = ["stm32h5/stm32h562", "device-selected", "rm0481"]
-stm32h563 = ["stm32h5/stm32h563", "device-selected", "rm0481"]
-stm32h573 = ["stm32h5/stm32h573", "device-selected", "rm0481"]
+stm32h563 = ["stm32h5/stm32h563", "device-selected", "rm0481", "sdmmc2", "ethernet"]
+stm32h573 = ["stm32h5/stm32h573", "device-selected", "rm0481", "otfdec", "sdmmc2", "ethernet"]
```
usbalbin, line 42: From what i can tell from Table 4 on page 114 of [rm0481](https://www.st.com/resource/en/reference_manual/rm0481-stm32h52333xx-stm32h56263xx-and-stm32h573xx-armbased-32bit-mcus-stmicroelectronics.pdf) STM32H523/533xx and STM32H562/563/573xx all have the OTFDEC1

src/rcc/rec.rs
```diff
@@ -504,69 +513,117 @@ peripheral_reset_and_enable_control! {
...
     ];
     #[cfg(feature = "rm0492")]
     AHB2, "" => [
-        Rng [kernel clk: Rng RNG ccipr5 "RNG"],
-        Adc [group clk: AdcDac(Variant) ADCDAC ccipr5 "ADC/DAC"],
         Dac12 [group clk: AdcDac]
     ];
+    #[cfg(feature = "rm0481")]
+    AHB2, "" => [
+        Gpioi,
```
usbalbin, line 526: Not sure what the plan is with STM32H523/533xx but they do not seem to support GPIOI even though they are rm0481 (see table 4, page 114)

src/rcc/rec.rs
```diff
@@ -504,69 +513,117 @@ peripheral_reset_and_enable_control! {
...
+    APB1L, "" => [
+        Uart8 [kernel clk: Uart8(Variant) USART ccipr1 "UART8"],
+        Uart7 [kernel clk: Uart7(Variant) USART ccipr1 "UART7"],
+        Cec [kernel clk: Cec(Variant) CEC ccipr5 "CEC"],
+        Usart11 [kernel clk: Usart11(Variant) USART ccipr2 "USART11"],
+        Usart10 [kernel clk: Usart10(Variant) USART ccipr1 "USART10"],
+        Usart6 [kernel clk: Usart6(Variant) USART ccipr1 "USART6"],
+        Uart5 [kernel clk: Uart5(Variant) USART ccipr1 "USART5"],
+        Uart4 [kernel clk: Uart4(Variant) USART ccipr1 "USART4"],
+        Tim14, Tim13, Tim12, Tim5, Tim4
```
usbalbin, line 577: Same as above with Tim14 and Tim13 (page 113)

src/rcc/rec.rs
```diff
@@ -504,69 +513,117 @@ peripheral_reset_and_enable_control! {
...
+    #[cfg(feature = "rm0481")]
     APB2, "" => [
-        Usb [kernel clk: Usb USB ccipr4 "USB"],
-        Usart1 [kernel clk: Usart1(Variant) USART ccipr1 "USART1"],
-        Spi1 [kernel clk: Spi1(Variant) SPI123 ccipr3 "SPI1"]
+        Sai2 [kernel clk: Sai2(Variant) SAI ccipr5 "SAI2"],
+        Sai1 [kernel clk: Sai1(Variant) SAI ccipr5 "SAI1"],
+        Spi6 [kernel clk: Spi6(Variant) SPI456 ccipr3 "SPI6"],
+        Spi4 [kernel clk: Spi4(Variant) SPI456 ccipr3 "SPI4"],
+        Tim17, Tim16, Tim15, Tim8
```
usbalbin, line 606: Same as above for all but Tim15 and Tim8 (page 112)

src/rcc/rec.rs
```diff
@@ -504,69 +513,117 @@ peripheral_reset_and_enable_control! {
...
     APB3, "" => [
-        I3c2 [kernel clk: I3c2(Variant) I3C ccipr4 "I3C2"],
-        LpTim1 [kernel clk: LpTim1(Variant) LPTIM ccipr2 "LPTIM1"],
-        LpUart1 [kernel clk: LpUart1(Variant) USART ccipr3 "LPUART1"]
+        Vref,
+        LpTim6 [kernel clk: LpTim6(Variant) LPTIM ccipr2 "LPTIM6"],
+        LpTim5 [kernel clk: LpTim5(Variant) LPTIM ccipr2 "LPTIM5"],
+        LpTim4 [kernel clk: LpTim4(Variant) LPTIM ccipr2 "LPTIM4"],
+        LpTim3 [kernel clk: LpTim3(Variant) LPTIM ccipr2 "LPTIM3"],
+        I2c4 [kernel clk: I2c4 I2C ccipr4 "I2C4"],
```
usbalbin, line 624: Same as above(page 111)

Cargo.toml
astapleton: The OFTDEC section in RM0481 (Section 37, page 1423) states that it is only relevant to STM32H533/STM32H573. It is also missing from the SVDs of 562/563.

src/rcc/rec.rs
astapleton: I was planning to handle the difference between GPIOs between H523/H533 and the other RM0481 peripherals in a follow up PR focused on GPIO definitions. I can add a feature then that distinguishes between GPIOs available for the two sets of processors.

astapleton: However, considering that there are similar concerns for other peripherals as you've highlighted below, I'll do it in this PR and do it in a consistent way.

src/rcc/rec.rs
astapleton: According to the STM32H523/STM32H533 datasheets, TIM15 and TIM8 are present. TIM16, and TIM17 are not

astapleton:
@usbalbin I've split peripheral definitions further based on your comments. PTAL.

usbalbin (approved):
I have not gone through the list again but looks good.

Even if there would be some error with the 523/533 I suppose it makes sense to fix that once that is added?

astapleton:
> Even if there would be some error with the 523/533 I suppose it makes sense to fix that once that is added?

Agreed!

Issue #34, opened 2025-01-20
Feature request - RNG


usbalbin:
Add support for the RNG peripheral which is available on all devices, also including H523 and H533.

astapleton:
@usbalbin Similar to #35 I don't have time to develop this in the short term, but am happy to help/review anything.

usbalbin:
Same as #35, me and/or @Wratox are happy to give this a try unless someone else is faster :)

PR #37, opened 2025-01-23
Add RNG

Wratox:
Added RNG as per #34, it compiles but I haven't been able to test it since I don't have access to a processor yet.


astapleton:
Thanks for the PR! Will you be getting hardware to be able to test this? I'll take more of a look in the next few days to a week, but my first inclination is that it would be nice to use generics instead of macros here. From my cursory look, it seems like it would be possible to create a generic RNG type that takes a numeric type as a parameter. That would probably require a trait bound that informs the compiler that the type implements whatever functionality is needed.

usbalbin:
We have ordered a Nucleo-H533 which should arrive any day now. The plan is to later make a custom pcb with the H523. So we will most likely shift our focus to #38 before we ca do any real tests with this

astapleton:
> We have ordered a Nucleo-H533 which should arrive any day now. The plan is to later make a custom pcb with the H523. So we will most likely shift our focus to #38 before we ca do any real tests with this

Great. I created https://github.com/stm32-rs/stm32h5xx-hal/issues/39 to track what we need to do to integrate STM32H523/STM32H533

src/rng.rs
```diff
@@ -0,0 +1,263 @@
...
+macro_rules! rng_core_large {
+    ($($type:ty),+) => {
+        $(
+            impl RngCore<$type> for Rng {
+                fn gen(&mut self) -> Result<$type, ErrorKind> {
+                    const WORDS: usize = mem::size_of::<$type>() / mem::size_of::<u32>();
+                    let mut res: $type = 0;
+
+                    for i in 0..WORDS {
+                        res |= (self.value()? as $type) << (i * (mem::size_of::<u32>() * 8))
```
astapleton, line 163:
```suggestion
                        res |= (self.value()? as $type) << (i * (u32::BITS as usize))
```

src/rng.rs
```diff
@@ -0,0 +1,263 @@
...
+            impl RngCore<$type> for Rng {
+                /// Returns a single element with random value, or error
+                fn gen(&mut self) -> Result<$type, ErrorKind> {
+                    let val = self.value()?;
+                    Ok(val as $type)
+                }
+
+                /// Fills buffer with random values, or return error
+                fn fill(&mut self, buffer: &mut [$type]) -> Result<(), ErrorKind> {
+                    const BATCH_SIZE: usize = 4 / mem::size_of::<$type>();
```
astapleton, line 133:
```suggestion
                    const BATCH_SIZE: usize = mem::size_of::<u32>() / mem::size_of::<$type>();
```

src/rng.rs
```diff
@@ -0,0 +1,263 @@
...
+            // This should never fail as long as no enum variant is equal to 0
+            .expect("Internal hal error")
+            .into()
+        })
+    }
+}
+
+#[cfg(feature = "rand")]
+#[cfg_attr(docsrs, doc(cfg(feature = "rand")))]
+impl rand_core::CryptoRng for Rng {}
```
astapleton, line 263: This marker trait is to indicate that this RNG is a cryptographically secure source of entropy (see https://docs.rs/rand_core/latest/rand_core/trait.CryptoRng.html). Looking at the user manual it suggests that the default configuration, which is what is being used here, is _not_ suitable for cryptographic use and refers to a NIST compatibile configuration as specified by AN4230. I don't think we should implement this marker trait with the current configuration, as it may lead to someone blindly using it as a cryptographic source and producing possibly insecure cryptographic material. I think we should (now or in the future) offer different configurations for cryptographic use and only implement this marker trait for a suitable configuration (while making no guarantees as to their suitability as that should be verified independently by consumers of this HAL)

src/rng.rs
```diff
@@ -0,0 +1,263 @@
...
+    fn fill(&mut self, dest: &mut [W]) -> Result<(), ErrorKind>;
+}
+
+pub struct Rng {
+    rb: RNG,
+}
+
+impl Rng {
+    /// Returns 32 bits of randomness, or error
+    pub fn value(&mut self) -> Result<u32, ErrorKind> {
```
astapleton, line 83: Can you provide a non-blocking version of this function (`value_nb`) that returns an `nb::Result` (see [nb](https://docs.rs/nb/latest/nb/) crate) indicating that it would block if the data is not ready, and then call the non-blocking function using the `nb::block!` macro here?

src/rng.rs
```diff
@@ -0,0 +1,263 @@
...
+        loop {
+            let status = self.rb.sr().read();
+            if status.cecs().bit() {
+                return Err(ErrorKind::ClockError);
+            }
+            if status.secs().bit() {
+                return Err(ErrorKind::SeedError);
```
astapleton, line 90: There's a seed error recovery process documented in the manual. Would you mind implementing that so when this error occurs, the recovery process can be run? As it stands, if this occurs the RNG would hang and be unrecoverable afterwards, except by using the PAC directly.

src/rng.rs
```diff
@@ -0,0 +1,263 @@
...
+        &self.rb
+    }
+
+    /// Returns a mutable reference to the inner peripheral
+    pub fn inner_mut(&mut self) -> &mut RNG {
+        &mut self.rb
+    }
+}
+
+impl core::iter::Iterator for Rng {
```
astapleton, line 113: neat 👍

astapleton (changes requested):
Thanks! I think this is a good start, but would like to see some seed error recovery, a non-blocking API and removal of the CryptoRng marker trait as mentioned in the comments. I know you brought this directly over from the H7 where it is already in use, but these feel like important changes to me.

src/rng.rs
astapleton: Note: `nb` is being brought in as a dependency in #27. There is also an open comment thread about it's use here: https://github.com/stm32-rs/stm32h5xx-hal/pull/27#discussion_r1908118667 (I'm hoping to get some closure on that soon!)

src/rng.rs
usbalbin: Do you want this implemented as its own method on the `Rng` type which the user then calls whenever they get an `ErrorKind::SeedError`?

src/rng.rs
usbalbin: @astapleton We have looked at the AN4230 which specifies the register values to use:

| Product       | RNG_CR         | HTCR | NSCR |
|------------------|-------------------|----------|---------|
| H503            | 0x00F00D00 | 0xAAC7 | default |
| H56x/H573 |  0x00F00E00 | 0x6A91 | 0x3AF66 |

It does not seem to mention the H523 and H533 parts eventhough they, as far as I can tell, both have NIST SP 800-90B(whatever that means) compliant RNG.


Either way, since the register values are given without, from what I have read so far, any motivation do you prefer
writing field by field like `w.field1().bits(1234).field2().bits(456)`
or simply write the entire blob to the register `w.bits(123456)`?

astapleton: > @astapleton We have looked at the AN4230 which specifies the register values to use:
> 
> Product	RNG_CR	HTCR	NSCR
> H503	0x00F00D00	0xAAC7	default
> H56x/H573	0x00F00E00	0x6A91	0x3AF66
> It does not seem to mention the H523 and H533 parts eventhough they, as far as I can tell, both have NIST SP 800-90B(whatever that means) compliant RNG.

I think let's be very clear about how the Rng is being initialized. Let's use separate functions to initialize the Rng in default vs ST's NIST configuration, and I think we should clearly label and the NIST init function (e.g. init_default() vs init_nist_st_an4230(), or something to that effect), and we should document that this uses the settings in AN4230, but have not performed the verification (buyer beware, users can/should do their own verification etc).

Given that there aren't settings documented for the H523/H533 parts yet, I think add the NIST init function to a separate extension trait and only implement it for the other processors until there are settings available.

> Either way, since the register values are given without, from what I have read so far, any motivation do you prefer writing field by field like `w.field1().bits(1234).field2().bits(456)` or simply write the entire blob to the register `w.bits(123456)`?

I'm not very particular about this. We're writing magic numbers anyway, so just add comments explaining what you're doing.

src/rng.rs
astapleton: So the general approach I've been taking is to expose non-blocking and blocking functions for each driver (you can see that in the I2C driver, but the SPI and USART drivers I have written and am intending to publish as PRs when the I2C driver lands follow then same pattern). This is intended to facilitate various use cases (interrupts vs spin-loop polling) and async extensions later.

For this case I can actually see 3 behaviours that are worth exposing: 
- Non-blocking calls ((e.g. `value_nb` + `recover_nb`) that return nb::Result and which do not automatically recover from the error
- Blocking calls (e.g. `value` + `recover`, which probably just call the above functions in a loop) that do not automatically recover from the error
- Blocking calls that automatically recover

The last is necessary for the module to be useful for the `rand_core` trait implementations, as they don't return `Result`s

src/rng.rs
```diff
@@ -0,0 +1,263 @@
...
+use core::mem;
+
+use crate::rcc::{rec, rec::RngClkSel};
+use crate::rcc::{CoreClocks, ResetEnable};
+use crate::stm32::RNG;
+use crate::time::Hertz;
+
+#[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum ErrorKind {
```
astapleton, line 18: nit: I've been using `Error` for Error enums

src/rng.rs
astapleton: Also, only the implementation that is initialized with`init_nist_st_an4230` should implement the `CryptoRng` trait, IMO

src/rng.rs
```diff
@@ -0,0 +1,389 @@
...
+/// Requires RNG to be disabled since some register values can only be written when RNGEN = 0
+pub trait RngNist {
+    fn rng_nist_st_an4230(self, prec: rec::Rng, clocks: &CoreClocks) -> Rng;
+}
+
+#[cfg(any(
+    feature = "stm32h562",
+    feature = "stm32h563",
+    feature = "stm32h573"
+))]
```
astapleton, line 73: There's a nist configuration for the STM32H503 isn't there?

src/rng.rs
```diff
@@ -0,0 +1,389 @@
...
+                .clear_bit()
+                .condrst()
+                .set_bit()
+        });
+
+        // Set health test control register values
+        self.htcr().write(|w| unsafe { w.bits(0xAAC7) });
+
+        // Set noise source control register
+        #[cfg(not(feature = "stm32h503"))] // Not available on H503
```
astapleton, line 140: Huh. I see this register is missing from the PAC, and the SVD, but it is documented in the reference manual. This is probably not a blocker, because the STM32H503 configurations for B, C require that the register stays at default, and if the register is not exposed via the PAC it is hard to modify. I'll submit a PR to stm32-rs to add it, but we don't need to depend on it to merge this, IMO. Can you leave a note in a comment to this effect here?

src/rng.rs
Wratox: Not sure how we want to do here, 
taken from rm0492:
> RNG configurations are described in Table 156. Note that only configuration A can be
certified NIST SP800-90B. Refer to Table 157 to select the best configuration for the
application.
• Configuration C can be used when configuration A is not flagged as certified

and according to an4230, stm32h503 configuration A is not certified.

Should we still use configuration A  for stm32503 or configuration C?
It's a bit unclear to me atleast.

src/rng.rs
```diff
@@ -0,0 +1,389 @@
...
+}
+
+#[cfg(any(
+    feature = "stm32h562",
+    feature = "stm32h563",
+    feature = "stm32h573"
+))]
+impl RngNist for RNG {
+    /// This uses the register values specified in AN4230 but have not
+    /// performed the verification (buyer beware, users can/should do their own verification)
```
astapleton, line 76: nit:
```suggestion
    /// Note:
    /// This uses the register values specified in AN4230 but verification 
    /// using this HAL has not been performed. Users can/should do their.
    ///  own verification or request documentation from ST directly.
```

src/rng.rs
astapleton: Oh interesting. I hadn't noticed that. Looking at AN4230, the way I understand it is that those are the recommended settings for NIST verification, but only the processor lines indicated were actually certified by ST, but that you _could_ certify using those settings. Maybe we should include it but note that ST recommends, but did not certify themselves with these settings?

Wratox: Sounds good to me 👍

src/rng.rs
astapleton: So the feedback I got on #27 was to not use `nb` and to not expose non-blocking APIs, as they're not that useful for composing any interrupt driven or asynchronous code, so I think ignore this request. I think if non-blocking calls are needed they can be added later.

src/rng.rs
astapleton: See comment above about leaving out the `_nb` calls.

src/rng.rs
```diff
@@ -0,0 +1,431 @@
...
+                .clear_bit()
+                .condrst()
+                .set_bit()
+        });
+
+        // Set health test control register values
+        self.htcr().write(|w| unsafe { w.bits(0xAAC7) });
+
+        // Set noise source control register
+        #[cfg(not(feature = "stm32h503"))] // Not available on H503
```
astapleton, line 202: Note that I modified the PAC to include this register on STM32H503 and it should be included in the latest release (`stm32h5` 0.16.0, released yesterday). Nothing to do on this PR, but we should integrate the new PAC and make this change afterwards.

src/rng.rs
```diff
@@ -0,0 +1,431 @@
...
+impl<MODE> Rng<MODE> {
+    /// Returns 32 bits of randomness, or error
+    pub fn value(&mut self) -> Result<u32, Error> {
+        loop {
+            let status = self.rb.sr().read();
+
+            if status.cecs().bit() {
+                return Err(Error::ClockError);
+            } else if status.secs().bit() {
+                return Err(Error::SeedError);
```
astapleton, line 256: So, from my understanding, we're configuring the RNG peripheral to automatically recover from seed error failures (no deactivation via the CR.ARDIS bit). Does that mean the consumer is expected to keep polling this function until it gets a valid value if it returns SeedError? If so, this should be documented somewhere. And should we be doing that in the iterator implementation?

astapleton (commented):
Generally looks good, but I think we need some clarification on the seed error recovery behavior (see inline comment).

src/rng.rs
```diff
@@ -278,10 +278,10 @@ impl<MODE> Rng<MODE> {
 }
 
 impl<MODE> core::iter::Iterator for Rng<MODE> {
-    type Item = u32;
+    type Item = Result<u32, Error>;
 
-    fn next(&mut self) -> Option<u32> {
-        self.value().ok()
+    fn next(&mut self) -> Option<Self::Item> {
+        Some(self.value())
```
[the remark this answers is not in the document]
usbalbin:
```rust
impl<MODE> core::iter::Iterator for Rng<MODE> {
    type Item = u32;

    fn next(&mut self) -> Option<Self::Item> {
        loop {
            match self.value() {
                Ok(x) => return Some(x),
                // We recover automatically from this, so try again
                Err(Error::SeedError) => (),
                Err(Error::ClockError) => return None,
            }
        }
    }
}
```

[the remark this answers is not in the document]
usbalbin: 32.3.7 of [rm0481](https://www.st.com/resource/en/reference_manual/rm0481-stm32h52333xx-stm32h56263xx-and-stm32h573xx-armbased-32bit-mcus-stmicroelectronics.pdf)

> In parallel to random number generation a health check block verifies the correct noise
source behavior and the frequency of the RNG source clock as detailed in this section.
Associated error state is also described.
Clock error detection
When the clock error detection is enabled (CED = 0) and if the RNG clock frequency is too
low, the RNG sets to 1 both the CEIS and CECS bits to indicate that a clock error occurred.
In this case, the application must check that the RNG clock is configured correctly (see
Section 32.3.6: RNG clocking) and then it must clear the CEIS bit interrupt flag. The CECS
bit is automatically cleared when the clocking condition is normal.
Note: The clock error has no impact on generated random numbers that is the application can still
read the RNG_DR register.
CEIS is set only when CECS is set to 1 by RNG.

--

>Is the clock error something we can check on init?

We do have the assert at line 52

```
// Otherwise clock checker will always flag an error
// See RM0481 Rev 2 Section 32.3.6
assert!(rng_clk > (hclk / 32), "RNG: Clock too slow");
```

I suppose we could also await `sr().read().drdy()` on init and check the clock status, that would probably catch any remaining clock setting errors.

However when it comes to check the flag for every value I am not sure. Are the pll's and clocks configs frozen on init or is there any chance of them changing? Also is the check also for the off chance that something fails in the hardware?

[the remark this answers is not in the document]
usbalbin:
```rust
impl<MODE> core::iter::Iterator for Rng<MODE> {
    type Item = u32;

    fn next(&mut self) -> Option<Self::Item> {
        loop {
            match self.value() {
                Ok(x) => return Some(x),
                // We recover automatically from a seed error, so try again
                Err(Error::SeedError) => (),
                Err(Error::ClockError) => {
                    #[cfg(feature = "log")]
                    log::warn!("RNG Clock error detected, retrying");

                    #[cfg(feature = "defmt")]
                    defmt::warn!("RNG Clock error detected, retrying");
                },
            }
        }
    }
}
```

do you want any message on seed error? If so, at what level?

src/rng.rs
astapleton: If you rebase on master now, you should be able to remove this line

src/rng.rs
```diff
@@ -0,0 +1,450 @@
...
+                #[cfg(feature = "defmt")]
+                defmt::warn!("RNG Clock error detected, retrying");
+
+                let sr = self.rb.sr();
+                // Give rng some time to recover from clock disturbance, this time seems to be about a handful of milliseconds
+                for _ in 0..100_000 {
+                    if sr.read().cecs().bit_is_clear() {
+                        continue 'outer;
+                    }
+                }
```
astapleton, line 262: Should we be spinning in a loop here hoping to recover?

astapleton (commented):
I think I've possibly confused the issue or wanted this API to do a bit too much. 

I've gone back and looked at the STM32H7 and STM32F4 HALs and they have much simpler error handling, which I think you had originally. Unless you feel that what we have now is better, how about we return to that (ie. like https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/rng.rs#L81-L93). And let's just end iteration in the Iterator implementation early when an error occurs, like you had before (and STM32H7 does: https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/rng.rs#L110-L116)

Apologies for driving this a bit off the rails.

src/rng.rs
```diff
@@ -0,0 +1,450 @@
...
+rng_core!(usize);
+
+// If usize is 64 bits, use the rng_core_large! impl
+#[cfg(target_pointer_width = "64")]
+rng_core_large!(usize);
+
+// rand_core
+#[cfg(feature = "rand")]
+#[cfg_attr(docsrs, doc(cfg(feature = "rand")))]
+impl<MODE> rand_core::RngCore for Rng<MODE> {
```
usbalbin, line 412: In stm32-rs/stm32g4xx-hal#224 TryCoreRng is implemented instead. Should we do the same?

src/rng.rs
```diff
@@ -0,0 +1,450 @@
...
+    };
+}
+
+// Only for types larger than 32 bits
+macro_rules! rng_core_large {
+    ($($type:ty),+) => {
+        $(
+            impl<MODE> RngCore<$type> for Rng<MODE> {
+                fn gen(&mut self) -> Result<$type, SeedError> {
+                    const WORDS: usize = mem::size_of::<$type>() / mem::size_of::<u32>();
```
usbalbin, line 345: Does this work with sizeof(T) < sizeof(u32)

src/rng.rs
astapleton: Considering it's fallible, this makes sense to me

PR #41, opened 2025-01-29
Add defmt

Wratox:


.cargo/config.toml
```diff
@@ -1,8 +1,9 @@
 [target.thumbv8m.main-none-eabihf]
-runner = 'arm-none-eabi-gdb'
+runner = 'probe-rs run --connect-under-reset'
```
Wratox, line 2: Is this something you want?

How to run defmt with probe-rs:

`DEFMT_LOG=trace cargo r --example blinky -- --chip stm32h533RETx`

astapleton: Yeah. The private fork we use is using probe-rs, so happy to have this change

Cargo.toml
```diff
@@ -83,4 +85,4 @@ opt-level = "s" # optimize for binary size
 
 [[example]]
 name = "blinky"
-required-features = ["stm32h503"]
+#required-features = ["stm32h503"]
```
astapleton, line 88: I added this requirement because the example was written for the H503 nucleo board. If its the same pin assignment on other nucleo boards we can remove the feature, or add different pin assignments for different boards, I guess.

astapleton:
Nice. Adding `defmt` was something I was hoping to get around to

Cargo.toml
usbalbin: [um3121](https://www.st.com/resource/en/user_manual/um3121-stm32h5-nucleo64-board-mb1814-stmicroelectronics.pdf) seems to cover 503 and 533 which both have PA5 as "aruino pin D13" which also is the led.

However for [um3115](https://www.st.com/resource/en/user_manual/um3115-stm32h5-nucleo144-board-mb1404-stmicroelectronics.pdf) Nucleo-H563ZI it is PB0 by default(unless solderbridges are changed)

astapleton (approved):

PR #45, opened 2025-02-28
i2c: Add target implementation

astapleton:
This extends the I2C driver to support target operation. It adds an I2cTarget type to represent the I2C operation as the target. 

Target operation can be operated in automatic ACKing mode, letting the hardware manage clock stretching and ACKing automatically, or it can be operated in manual ACK control mode, which lets the user of the driver determine when to ACK the end of a transaction. Manual ACK control allows for clock stretching while data is being processed or retrieved before ACK'ing the final byte of a transfer and initiating either the end of a transfer or a repeat start to read the processed/retrieved data. This implementation does not currently allow for ACKing individual or a subset of bytes manually, but extending the interface to do so can build upon what is implemented so far. 

This also adds role switching to the driver, enabling both controller and target implementation through choosing the appropriate initialization functions.

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -1,5 +1,11 @@
 //! Inter Integrated Circuit (I2C)
 //!
+//! This module provides I2C functionality as both a Controller and Target, and
+//! supports multi-controller topologies through role-switching between
+//! Controller and Target. The Controller implementation is exposed via the I2c
+//! struct, while the Target implementation is exposed via the I2cTarget
+//! struct.
+//!
 //! # Terminology
 //! This uses the updated Controller/Target terminology that replaces Master/Slave, as of v7 of the
 //! I2C Spec. See https://www.nxp.com/docs/en/user-guide/UM10204.pdf
@@ -43,10 +49,103 @@
 //! i2c.write_read(0x18, &write, &mut read)?;
 //! ```
 //!
+//! ## Target
+//! In the simplest case, the I2cTarget can be initialized from the device peripheral
+//! and GPIO pins:
+//!
+//! ```
+//! let dp = ...;            // Device peripherals
+//! let (scl, sda) = ...;    // GPIO pins
+//! let own_addr = ...;      // Primary address for target operation
+//!
+//! let mut i2c_target = dp.I2C1.i2c_target_only(
+//!     (scl, sda),
+//!     own_addr,
+//!     ccdr.peripheral.I2C1,
+//! );
+//! ```
+//!
+//! The target (or controller when role switching is allowed) must be
+//! instructed to respond to specific TargetListenEvents using the
+//! `enable_target_event` function:
+//!
+//! ```
+//! i2c_target.enable_target_event(TargetListenEvent::PrimaryAddress);
+//! ```
+//!
+//! Use the `get_target_event_nb` (non-blocking) or `wait_for_event` (blocking)
+//! functions to get target events to which to respond. Then use the provided
+//! `read`/`write`` (blocking) or `read_nb`/`write_nb`(non-blocking) functions
+//! to respond to the controller:
+//!
+//! ```
+//! let mut buffer = [0u8; 10];
+//! match i2c.wait_for_event().unwrap() {
+//!     TargetEvent::Read { address: _ } => i2c.write(...),
+//!     TargetEvent::Write { address: _ } => i2c.read(&mut buffer),
+//!     TargetEvent::Stop => Ok(0),
+//! };
+//! ```
+//!
+//! ### Manual ACK control in receive mode
+//!
+//! To gain control over ACK'ing received data when operating as a target, convert to
+//! manual ACK control with:
+//!
+//! ```
+//! let mut i2c_target_manual_ack = i2c_target.with_manual_ack_control();
+//! ```
+//!
+//! This provides the methods `I2cTarget::ack_transfer` and
+//! `I2cTarget::nack_transfer` to end a transfer of an expected number of bytes
+//! with an ACK or NACK, respectively:
+//!
+//! ```
+//! let mut buf = [0u8; 10];
+//! i2c_target_manual_ack.read(&mut buf)?;  // Read 10 bytes, ACK each byte except the last.
+//! // Check something
+//! if good {
+//!     i2c_target_manual_ack.ack_transfer()
+//! } else {
+//!     i2c_target_manual_ack.nack_transfer()
+//! }
+//! ```
+//!
+//! ## Switching between target and controller
+//!
+//! Use the provided initialization functions to create an I2C driver that can switch
+//! between controller and target operation:
+//!
+//! ```
+//! let dp = ...;            // Device peripherals
+//! let (scl, sda) = ...;    // GPIO pins
+//! let own_addr = ...;      // Primary address for target operation
+//!
+//! let mut i2c = dp.I2C1.i2c_controller_target(
+//!     (scl, sda),
+//!     own_addr,
+//!     ccdr.peripheral.I2C1,
+//! );
+//! ```
+//!
+//! To switch operating modes, use the provided functions:
+//!
+//! ```
+//! let i2c_target = i2c.to_target();
+//! ...
+//! let i2c = i2c_target.to_controller();
+//! ```
+//!
+//! These functions are only available on driver instances created with the
+//! above initialization function.
+//!
 //! # Examples
 //!
 //! - [I2C controller simple example](https://github.com/stm32-rs/stm32h5xx-hal/blob/master/examples/i2c.rs)
+//! - [I2C Target simple example](https://github.com/stm32-rs/stm32h5xx-hal/blob/master/examples/i2c_target.rs)
+//! - [I2C Target with manual ack control](https://github.com/stm32-rs/stm32h5xx-hal/blob/master/examples/i2c_target_manual_ack.rs)
 
+use core::marker::PhantomData;
 use core::ops::{Deref, DerefMut};
 
 use crate::rcc::{CoreClocks, ResetEnable};
@@ -56,6 +155,9 @@ use crate::stm32::{i2c1, i2c1::isr::ISRrs};
 type Isr = stm32h5::R<ISRrs>;
 
 use crate::time::Hertz;
+
+pub mod config;
+pub use config::TargetConfig;
 
 mod hal;
 mod i2c_def;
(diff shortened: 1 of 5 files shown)


src/i2c.rs
```diff
@@ -108,6 +210,48 @@ pub enum Error {
...
+/// Target Event.
+///
+/// This encapsulates a transaction event that occurs when listening in Target
+/// operation. A TargetWrite event indicates that the Controller wants to read
+/// data from the Target (I2C read operation). A TargetRead event indicates
+/// that the Controller wants to write data to the Target (I2C write
+/// operation). A Stop event indicates that a Stop condition was received and
+/// the transaction has been completed.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum TargetEvent {
```
ryan-summers, line 234: These definitions are pretty confusing. Wouldn't a `TargetRead` indicates that the controller wants to read data from the target?

src/i2c.rs
```diff
@@ -108,6 +210,48 @@ pub enum Error {
...
+}
+
+/// Target Listen Event
+///
+/// Indicates what listen events are responded to. A target can respond to one
+/// of or all of a general call address, the primary address configured or a
+/// secondary address.
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub enum TargetListenEvent {
+    GeneralCall,
```
ryan-summers, line 252: I'd document what these are supposed to be. I get primary/secondary address, but GeneralCall is pretty opaque.

src/i2c.rs
```diff
@@ -718,6 +1077,536 @@ impl<I2C: Instance> I2c<I2C> {
...
+                bytes[i - 1]
+            } else {
+                0
+            };
+            match self.write_byte(data) {
+                Ok(()) => {}
+                Err(Error::NotAcknowledge) => return Ok(i),
+                Err(error) => return Err(error),
+            }
+        }
```
ryan-summers, line 1123: I'd recommend using an iterator over `bytes` instead. `for (byte, i) in bytes.iter().enumerate()` instead. That prevents manually managing the index.

src/i2c.rs
```diff
@@ -718,6 +1077,536 @@ impl<I2C: Instance> I2c<I2C> {
...
+                // Flush the contents of TXDR without writing it to the bus. Doing so ensures that
+                // spurious data is not written to the bus and that the clock remains stretched when
+                // the ADDR flag is cleared and until data is written to TXDR (in the case of a
+                // target write operation)
+                self.i2c.isr().write(|w| w.txe().set_bit());
+            }
+            self.i2c.icr().write(|w| w.addrcf().clear());
+            let address = isr.addcode().bits();
+            if isr.dir().is_read() {
+                // TODO: Translate address into 10-bit stored address if relevant
```
ryan-summers, line 1142: I'd open an issue for this just for tracking purposes

src/i2c.rs
```diff
@@ -718,6 +1077,536 @@ impl<I2C: Instance> I2c<I2C> {
...
+    /// If the transaction was stopped, TargetEvent::Stop will be returned.
+    ///
+    /// If an error occurs while waiting for an event, this will be returned.
+    ///
+    /// Note: The TargetEvent type that is indicated is indicated from the
+    /// perspective of the Bus controller, so if a Read event is received,
+    /// then the Target should respond with a write, and vice versa.
+    pub fn wait_for_event(&mut self) -> Result<TargetEvent, Error> {
+        loop {
+            if let Some(event) = self.get_target_event()? {
```
ryan-summers, line 1373: Can this be put in the base trait implementation as a default? It seems to be copied across the impls

src/i2c.rs
astapleton: Ha. I actually had these the other way round, simply called `TargetEvent::Read` (controller wants to read, target must execute a write) and `TargetEvent::Write` (controller wants to write, target must execute a read), but I thought that was confusing, so I tried to disambiguate it by indicating that a TargetRead event is in response to controller write. Guess they're both a bit confusing, so I'll go back to the old iteration.

src/i2c.rs
astapleton: The manual accounting here is to write valid data while the whole buffer has not been read, and to just write out zeroes afterwards (controller keeps clocking bytes out after the buffer is empty). Guess I could use an iterator and manually calling the next method and replace a `None` with 0.

src/i2c.rs
```diff
@@ -606,6 +989,12 @@ impl<I2C: Instance> I2c<I2C> {
     ///
     /// See the sequence diagrams in Figures 390 & 393 of RM0492 Rev 2 for more.
     fn reload(&self, length: usize, direction: Direction, stop: Stop) {
+        assert!(
+            length <= u8::MAX as usize,
+            "I2C max transfer size per reload = {}",
+            u8::MAX
+        );
```
astapleton, line 996: I added these bounds checks in the controller start logic because I realized that the controller implementation does not use NBYTES reloads to write buffers longer than 255 bytes out the buffer, but also doesn't bounds check on repeat starts or reloads, leading to wrap around bugs on the value written to the NBYTES field and transactions that were unexpectedly short.

astapleton:
@ryan-summers mind taking another look at this? I believe I've addressed everything

src/i2c.rs
```diff
@@ -378,24 +665,70 @@ impl<I2C: Instance> I2c<I2C> {
...
+        self.inner.i2c
     }
+}
+
+/// Target implementation
+impl<I2C: Instance, A, R> I2cTarget<I2C, A, R> {
+    fn new(
+        i2c: I2C,
+        target_config: impl Into<TargetConfig>,
+        rec: <I2C as Instance>::Rec,
```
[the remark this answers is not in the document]
astapleton: It was. Don't know why I had that.

src/i2c.rs
```diff
@@ -718,6 +1107,501 @@ impl<I2C: Instance> I2c<I2C> {
...
+        // zeroes until the controller stops the transaction.
+        // This creates an iterator that will return zeroes when the buffer is exhausted.
+        for (i, data) in bytes.iter().chain(iter::repeat(&0)).enumerate() {
+            match self.write_byte(*data) {
+                Ok(()) => {}
+                Err(Error::NotAcknowledge) => {
+                    // If we receive a NACK it will be on the FIFO write subsequent to the last byte
+                    // actually written to the bus. If we start writing zeroes out, we only want to
+                    // indicate how many bytes from the buffer we wrote.
+                    return Ok(core::cmp::min(i - 1, bytes.len()));
```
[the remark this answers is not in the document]
usbalbin: or do we need a saturating sub?

src/i2c.rs
```diff
@@ -718,6 +1107,501 @@ impl<I2C: Instance> I2c<I2C> {
...
+    pub fn write(&mut self, bytes: &[u8]) -> Result<usize, Error> {
+        TargetAckMode::write(self, bytes)
+    }
+
+    /// Waits for a controller write event and reads the data written by the
+    /// controller to the buffer provided. This will read until the buffer is
+    /// full, or until the transfer is stopped by the controller, whichever
+    /// comes first.
+    ///
+    /// If the a bus error occurs or a stop condition is received, an error
```
usbalbin, line 1440:
```suggestion
    /// If a bus error occurs or a stop condition is received, an error
```

src/i2c.rs
```diff
@@ -333,7 +598,28 @@ fn calc_timing_params(ker_ck: u32, target_freq: u32) -> (u8, u8, u8, u8, u8) {
...
+        }
+        AddressMode::AddressMode10bit => {
+            w.oa1().set(config.own_address).oa1mode().bit10()
+        }
+    });
+
+    if let Some(secondary_address) = config.secondary_address {
+        i2c.oar2().write(|w| {
+            w.oa2().set(secondary_address);
+            if let Some(mask) = config.secondary_address_mask_bits {
```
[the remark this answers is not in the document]
astapleton: The mask allows for the target to respond to a wider range of addresses (ie. an address of 0x20 and mask of 0x1F will let the target watch for transmissions on any address from 0x20 to 0x3F). This is an STM32 specific feature, not a standard feature of I2C, by the way.

src/i2c.rs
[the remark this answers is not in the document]
astapleton: Ah, good catch. We should use a saturating sub here

src/i2c.rs
```diff
@@ -718,6 +1107,501 @@ impl<I2C: Instance> I2c<I2C> {
...
+impl<I2C: Instance, R> Targetable for I2cTarget<I2C, R> {
+    fn enable_listen_event(&mut self, event: TargetListenEvent) {
+        self.inner.enable_target_event(event)
+    }
+
+    fn disable_listen_event(&mut self, event: TargetListenEvent) {
+        self.inner.disable_target_event(event)
+    }
+
+    fn configure_target(&mut self, config: impl Into<TargetConfig>) {
```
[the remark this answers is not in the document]
astapleton: ah, In a previous revision I didn't include the bus frequency in the target config, so I had an implementation of u8 for Into<TargetConfig> so you could simply provide the address. Unfortunately, one needs to provide the frequency to set up timing parameters correctly, and providing a tuple was less ergonomic, so I just removed that implementation. And I will change to just use a TargetConfig here.

src/i2c.rs
```diff
@@ -108,6 +211,55 @@ pub enum Error {
...
+    /// Target operation only:
+    /// Indicates that a stop or repeat start was received while reading, or
+    /// while explicitly waiting for a controller read or write event.
+    TransferStopped,
+    /// Target operation only:
+    /// While waiting for a controller read event, a write event was received.
+    ControllerExpectedWrite,
+    /// Target operation only:
+    /// While waiting for a controller write event, a read event was received.
+    ControllerExpectedRead,
```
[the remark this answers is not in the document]
astapleton: My latest commit adds a `TargetError` enum and a `TargetError` for `Error`. Does that look better to you?

src/i2c.rs
```diff
@@ -108,6 +213,58 @@ pub enum Error {
     Arbitration,
     /// NACK received
     NotAcknowledge,
+    /// Target error occurred. Can be ignored during controller operation.
+    TargetError(TargetError),
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum TargetError {
```
usbalbin, line 221:
```suggestion
#[derive(Copy, Clone, Debug, PartialEq, Eq)]
#[cfg_attr(feature = "defmt", derive(defmt::Format))]
pub enum TargetError {
```

usbalbin (approved):
Very nice as far as I can tell! :)

PR #54, opened 2025-05-30
Fix bad offsets

usbalbin:
Found out that the offset for setting the pull up/down resistor is wrong during #52

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -316,7 +316,7 @@ where
 {
     /// Set pin speed
     pub fn set_speed(&mut self, speed: Speed) {
-        let offset = 2 * { N };
+        let offset = N;
 
         unsafe {
             (*Gpio::<P>::ptr())
@@ -338,7 +338,7 @@ where
 {
     /// Set the internal pull-up and pull-down resistor
     pub fn set_internal_resistor(&mut self, resistor: Pull) {
-        let offset = 2 * { N };
+        let offset = N;
         let value = resistor as u8;
         unsafe {
             (*Gpio::<P>::ptr())


astapleton:
Good catch! This looks like it comes from #28 where we updated to the new PAC with offsets baked in to the PAC calls. Can you do a pass over all the register accesses? It looks like the mode register accesses are also wrong.

usbalbin:
Here is a branch with the defmt changes, extra test and the fix https://github.com/stm32-rs/stm32h5xx-hal/compare/master...usbalbin:tests-w-defmt?expand=1 not sure what happened to the history, must have messed up some rebase. Anyways, I have to go now.

Here is the result:

```
cargo test --features stm32h533,defmt --tests -- --chip stm32h533RETx
   Compiling stm32h5xx-hal v0.0.0 (/home/albin/stm32h5xx-hal)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 3.33s
     Running tests/nucleo-h533.rs (target/thumbv8m.main-none-eabihf/debug/deps/nucleo_h533-5bdae1dc6f4190f1)
      Erasing ✔ 100% [####################] 104.00 KiB @ 663.37 KiB/s (took 0s)
  Programming ✔ 100% [####################] 102.00 KiB @  82.23 KiB/s (took 1s)
Finished in 1.40s

running 4 tests
test tests::gpio_settings   ... ok
test tests::gpio_resistors  ... ok
test tests::gpio_push_pull  ... ok
test tests::gpio_open_drain ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.98s
```

usbalbin:
So from what I have tested, #54 is all that is needed

oshaughnessya (approved):

astapleton (approved):

PR #55, opened 2025-05-31
Fix OSPEEDR and PUPDR indexing

Javier-varez:
The field width of 2 was being applied twice in these two registers, because the PAC already accounts for it, so there is no need to apply it externally.

Thankfully the PAC triggers and out-of-bounds access to an array when this happens in debug builds, so it can be easily caught.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -316,12 +316,10 @@ where
 {
     /// Set pin speed
     pub fn set_speed(&mut self, speed: Speed) {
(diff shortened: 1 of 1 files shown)


oshaughnessya:
There seems to be overlapping work, take a look at #54

Javier-varez:
Right, I didn't notice a fix had been recently submitted, will close this PR. Thank you!

PR #59, opened 2025-06-12
Add timer

tytouf:
Copy code from stm32h7 to implement basic timer on stm32h5.
Add an example using Timer2: blinky_timer


astapleton (changes requested):
Thanks for the PR! The main changes I think this needs:
-  remove the dependency on embedded-hal 0.2 (see comments), and replace the functionality that the trait implementations provided
- remove references of the stm32h7xx repository (unless intentional)

astapleton:
Closing this in favor of #76

Issue #50, opened 2025-05-27
Add self tests


usbalbin:
What do you think about adding some sort of self tests that can be run on a nucleo board as done [here](https://github.com/stm32-rs/stm32g4xx-hal/blob/main/tests/nucleo-g474.rs)?

astapleton:
That looks really cool. I think that would be a good addition. Do you want to work on that? I've got a bit of a backlog at the moment...

usbalbin:
Sure! I have some stuff to do first but then yes. I might try to clean up the linked one a bit first and then at some point try to move it here.

oshaughnessya:
I know I have looked into [embedded-test](https://github.com/probe-rs/embedded-test) in the past and prefer it.

usbalbin:
I have cleaned up the above a bit(still a lot of room for improvement), removed all tests for missing peripherals and ported it to this crate in #52. Turns out two gpio tests were the only tests I could keep. Will probably add some more gpio related.

Do you have any ideas for more things to test for? And if so then how? :)

usbalbin:
I do not know about you, but think there is quite a lot of value in having the vast majority, if not all of the test, be possible to run on a simple nucleo-503/533 with modifications that can be done in minutes with no tools just moving around the jumpers once. This would make it easier to get started for people.

astapleton:
> I do not know about you, but think there is quite a lot of value in having the vast majority, if not all of the test, be possible to run on a simple nucleo-503/533 with modifications that can be done in minutes with no tools just moving around the jumpers once. This would make it easier to get started for people.

Agreed. The H503 nucleo board is unfortunately very barebones, so there's not much to test against (for e.g. I2C), but you can route the SPI back to itself at least.

oshaughnessya:
CAN could probably be tested by putting it into loop back mode, sending and then receiving a message.

astapleton:
> SCK could perhaps be used for SPI to check the frequency, assuming it is set low enough to measure?

With the timer peripheral? Or with a logic analyzer/scope?

usbalbin:
While a lot better than nothing, I think there are probably quite o lot of errors that might still pass by unnoticed in a loopback test for things like UART. For example stop bits, baud rate etc. Since the errors would likely cancel out if the RX-part has the same error.

[...]

PR #61, opened 2025-07-10
dwt: copy DWT management and profiling features from stm32f4xx-hal (#50)

astapleton:
The DWT unit allows cycle count based profiling/tracing for performance testing. The [module in the
stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/dwt.rs) is a useful tool for managing this functionality, so I'm copying it verbatim (with very minor changes for imports) here.


usbalbin (approved):
LGTM

PR #62, opened 2025-07-11
spi: Add slave mode operation

### LINKED ISSUE #36: Feature request - SPI
Add support for the SPIx peripherals.

Might try to add this myself, but is there any reason to not go straight for embedded_hal v1.0.0? 
If so do you think [https://github.com/stm32-rs/stm32h7xx-hal/pull/476](https://github.com/stm32-rs/stm32h7xx-hal/pull/476) is a good place to start?

astapleton:
This adds slave operation to the SPI driver. The driver exposes the same set of APIs as those provided for master operation, with the exception that the blocking APIs are defined by the driver instead of the `embedded-hal traits`, which do not currently support slave operation.

This closes #36.


src/spi.rs
```diff
@@ -458,6 +587,78 @@ impl<SPI: Instance, W: Word> Spi<SPI, W> {
...
+                .comm()
+                .variant(config.communication_mode.into())
+        });
+
+        self
+    }
+
+    fn hardware_cs_enabled(&self) -> bool {
+        // SSM bit indicates software control of SS input is enabled if it is set
+        self.inner.spi.cfg2().read().ssm().is_disabled()
```
usbalbin, line 658: Just double checking *Not* SW means HW?

astapleton: It does, yeah:
Screenshot 2025-08-02 at 11 10 09

I can clarify that comment

usbalbin (approved):
Nice!

LGTM now :)

PR #63, opened 2025-07-20
USB

usbalbin:


examples/usb_serial.rs
```diff
@@ -0,0 +1,100 @@
...
+
+                if let Ok(s) = str::from_utf8(&buf[0..count]) {
+                    info!("{:?}", s);
+                } else {
+                    info!("{:?}", &buf[0..count]);
+                }
+
+                // Echo back in upper case
+                for c in buf[0..count].iter_mut() {
+                    if 0x61 <= *c && *c <= 0x7a {
```
usbalbin, line 80: Any ideas for solving this?

```
error: current MSRV (Minimum Supported Rust Version) is `1.78.0` but this item is stable since `1.87.0`
  --> examples/usb_serial.rs:72:32
   |
72 |                 if let Ok(s) = str::from_utf8(&buf[0..count]) {
   |                                ^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#incompatible_msrv
note: the lint level is defined here
  --> examples/usb_serial.rs:2:9
```

astapleton: We could bump up the MSRV? We haven't published a crate yet...

usbalbin: Any suggestions for a MSRV?

usbalbin:
Note that this is only tested on a Nucleo-H533RE

src/usb.rs
```diff
@@ -0,0 +1,116 @@
...
+                // Set bit to confirm that USB supply level is good
+                pwr.usbscr().modify(|_, w| w.usb33sv().set_bit());
+            }
+
+            // Enable USB peripheral
+            rcc.apb2enr().modify(|_, w| w.usben().set_bit());
+
+            // Reset USB peripheral
+            rcc.apb2rstr().modify(|_, w| w.usbrst().set_bit());
+            rcc.apb2rstr().modify(|_, w| w.usbrst().clear_bit());
```
astapleton, line 106: How about:
```
rec::Usb { _marker: PhantomData }.reset().enable();
```

src/usb.rs
```diff
@@ -0,0 +1,116 @@
...
+    const REGISTERS: *const () = USB::ptr().cast::<()>();
+    const DP_PULL_UP_FEATURE: bool = true;
+    const EP_MEMORY: *const () = 0x4001_6400 as _;
+    const EP_MEMORY_SIZE: usize = 2048;
+    const EP_MEMORY_ACCESS: stm32_usbd::MemoryAccess =
+        stm32_usbd::MemoryAccess::Word32x1;
+
+    fn enable() {
+        cortex_m::interrupt::free(|_| {
+            let rcc = unsafe { &*stm32::RCC::ptr() };
```
astapleton, line 85: nit: move this closer to where it's first used.

src/usb.rs
usbalbin: Nice, thanks

usbalbin:
Note: I have not tested this since b6c7432

src/usb.rs
```diff
@@ -0,0 +1,118 @@
...
+use crate::stm32::{self, USB};
+use core::fmt;
+use core::marker::PhantomData;
+use stm32_usbd::UsbPeripheral;
+
+/// Type for pin that can be the "D-" pin for the USB peripheral
+pub type DmPin = PA11<gpio::Alternate<10>>;
+
+/// Type for pin that can be the "D+" pin for the USB peripheral
+pub type DpPin = PA12<gpio::Alternate<10>>;
```
astapleton, line 20: is this consistent across all parts?

src/usb.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+pub struct UsbDevice {
+    /// USB register block
+    _usb: USB,
+    /// Data negative pin
+    pin_dm: DmPin,
+    /// Data positive pin
+    pin_dp: DpPin,
```
astapleton, line 46: Do we need to hang on to these pins? Doesn't looks like they get used anywhere. Could just consume them in the new function so they can't get used anywhere else (e.g. SPI driver does this).

src/usb.rs
```diff
@@ -0,0 +1,118 @@
...
+
+impl fmt::Debug for UsbDevice {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Peripheral")
+            .field("usb", &"USB")
+            .field("pin_dm", &self.pin_dm)
+            .field("pin_dp", &self.pin_dp)
+            .finish()
+    }
+}
```
astapleton, line 69: this is out of date. how about just using the `Debug` derive for this?

src/usb.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// Data positive pin
+    pin_dp: DpPin,
+}
+
+#[cfg(feature = "defmt")]
+impl defmt::Format for UsbDevice {
+    fn format(&self, f: defmt::Formatter) {
+        defmt::write!(
+            f,
+            "Peripheral {{ usb: USB, pin_dm: {}, pin_dp: {}}}",
```
astapleton, line 54: Name is out of date

astapleton (commented):
A few more comments

src/usb.rs
usbalbin: We actually don not use the pins for anything once they are put into the struct. As in we have no `release` method. Should I remove them from the struct?

src/usb.rs
usbalbin: I'll check..

astapleton (approved):
Awesome. Thanks for doing this! Looks good to me.

PR #64, opened 2025-07-31
gpdma: add GPDMA driver

astapleton:
This adds support for one-shot DMA transfers for all supported transfer types with linear addressing. It does not yet support linked list buffer loading, or support 2D addressing.

Memory-to-memory transfers were quite thoroughly tested by the provided example, memory-to-peripheral and peripheral-to-memory transfers were tested with the SPI peripheral (working on a test branch). Peripheral-to-peripheral transfers are assumed to work given their similarity to memory-to-peripheral/peripheral-to-memory transfers, but will be properly tested at a later stage.

The driver includes helper structs for peripheral transfers in the `gpdma::periph` module which handle the common operations for setting up one-directional and full-duplex transfers using one, or two channels, respectively.

Transfer configuration is performed via `DmaConfig`. Additionally, the `transform` module includes support for configuring the data transformation pipeline supported by the GPDMA peripheral, only allowing relevant options to be supported based on the source and destination word sizes.

### DIFF
--- a/examples/dma.rs
+++ b/examples/dma.rs
@@ -7,51 +7,11 @@ mod utilities;
 use cortex_m::singleton;
 use cortex_m_rt::entry;
 use cortex_m_semihosting::debug;
-use embedded_dma::{ReadBuffer, WriteBuffer};
 use stm32h5xx_hal::{
-    gpdma::{config::transform::*, DmaConfig, DmaTransfer, Word},
+    gpdma::{config::transform::*, DmaConfig, DmaTransfer},
     pac,
     prelude::*,
 };
-
-// Buffer is used to manage a reference to a static buffer returned by the cortex_m::singleton!
-// macro and which can be with the DmaTransfer API (which requires passing ReadBuffer and
-// WriteBuffer implementations by value) and then used to access the buffer after the transfer has
-// completed.
-struct Buffer<T: Word + 'static, const N: usize> {
-    data: &'static mut [T; N],
-}
-
-impl<T, const N: usize> Buffer<T, N>
-where
-    T: Word + 'static,
-{
-    fn new(data: &'static mut [T; N]) -> Self {
-        Self { data }
-    }
-}
-
-unsafe impl<T, const N: usize> ReadBuffer for &Buffer<T, N>
-where
-    T: Word + 'static,
-{
-    type Word = T;
-
-    unsafe fn read_buffer(&self) -> (*const Self::Word, usize) {
-        (self.data.as_ptr(), N)
-    }
-}
-
-unsafe impl<T, const N: usize> WriteBuffer for &mut Buffer<T, N>
-where
-    T: Word + 'static,
-{
-    type Word = T;
-
-    unsafe fn write_buffer(&mut self) -> (*mut Self::Word, usize) {
-        (self.data.as_mut_ptr(), N)
-    }
-}
 
 #[entry]
 fn main() -> ! {
@@ -72,26 +32,20 @@ fn main() -> ! {
     let src =
         singleton!(: [u8; 40] = core::array::from_fn(|i| i as u8)).unwrap();
 
-    let src_buffer = Buffer::new(src);
     let dest = singleton!(: [u8; 40] = [0u8; 40]).unwrap();
-    let mut dest_buffer = Buffer::new(dest);
 
     let mut channel = channels.0;
     let config = DmaConfig::new();
-    let mut transfer = DmaTransfer::memory_to_memory(
-        config,
-        &mut channel,
-        &src_buffer,
-        &mut dest_buffer,
-    );
+    let mut transfer =
+        DmaTransfer::memory_to_memory(config, &mut channel, src, dest);
     transfer.start().unwrap();
     transfer.wait_for_transfer_complete().unwrap();
-    assert_eq!(src_buffer.data, dest_buffer.data);
+    let (src, dest) = transfer.free();
+    assert_eq!(src, dest);
 
     log::info!("u32 to u32 with data transform");
     let src = singleton!(: [u32; 10] = [0x12345678u32; 10]).unwrap();
     let dest = singleton!(: [u32; 10] = [0u32; 10]).unwrap();
-    let mut dest_buffer = Buffer::new(dest);
 
     let config = DmaConfig::new().with_data_transform(
         DataTransform::builder()
@@ -99,62 +53,52 @@ fn main() -> ! {
             .swap_destination_half_word_byte_order(),
     );
 
-    let mut transfer = DmaTransfer::memory_to_memory(
-        config,
-        &mut channel,
-        src,
-        &mut dest_buffer,
-    );
+    let mut transfer =
+        DmaTransfer::memory_to_memory(config, &mut channel, src, dest);
 
     transfer.start().unwrap();
     transfer.wait_for_transfer_complete().unwrap();
+    let (_, dest) = transfer.free();
+
     let expected = [0x78563412; 10];
-    assert_eq!(expected, *dest_buffer.data);
+    assert_eq!(expected, *dest);
 
     log::info!("u32 to u16 with truncate");
     let src = singleton!(: [u32; 10] = [0x12345678u32; 10]).unwrap();
     let dest = singleton!(: [u16; 20] = [0u16; 20]).unwrap();
-    let mut dest_buffer = Buffer::new(dest);
 
     let config = DmaConfig::new().with_data_transform(
         DataTransform::builder().left_align_right_truncate(),
     );
-    let mut transfer = DmaTransfer::memory_to_memory(
-        config,
-        &mut channel,
-        src,
-        &mut dest_buffer,
-    );
+    let mut transfer =
+        DmaTransfer::memory_to_memory(config, &mut channel, src, dest);
 
     transfer.start().unwrap();
     transfer.wait_for_transfer_complete().unwrap();
+    let (_, dest) = transfer.free();
+
     let expected = [0x1234; 10];
-    assert_eq!(expected, (*dest_buffer.data)[0..10]);
+    assert_eq!(expected, (*dest)[0..10]);
 
     log::info!("u32 to u8 with unpack");
     let src = singleton!(: [u32; 10] = [0x12345678u32; 10]).unwrap();
     let dest = singleton!(: [u8; 40] = [0u8; 40]).unwrap();
-    let mut dest_buffer = Buffer::new(dest);
 
     let config =
         DmaConfig::new().with_data_transform(DataTransform::builder().unpack());
-    let mut transfer = DmaTransfer::memory_to_memory(
-        config,
-        &mut channel,
-        src,
-        &mut dest_buffer,
-    );
+    let mut transfer =
+        DmaTransfer::memory_to_memory(config, &mut channel, src, dest);
 
     transfer.start().unwrap();
     transfer.wait_for_transfer_complete().unwrap();
+    let (_, dest) = transfer.free();
     let expected = [0x78, 0x56, 0x34, 0x12];
-    assert_eq!(expected, (*dest_buffer.data)[0..4]);
-    assert_eq!(expected, (*dest_buffer.data)[36..40]);
+    assert_eq!(expected, (*dest)[0..4]);
+    assert_eq!(expected, (*dest)[36..40]);
 
     log::info!("u8 to u32 with pack");
     let src = singleton!(: [u8; 40] = [0u8; 40]).unwrap();
     let dest = singleton!(: [u32; 10] = [0u32; 10]).unwrap();
-    let mut dest_buffer = Buffer::new(dest);
 
     for chunk in src.chunks_mut(4) {
         chunk.copy_from_slice(&[0x78, 0x56, 0x34, 0x12]);
@@ -162,18 +106,16 @@ fn main() -> ! {
 
     let config =
         DmaConfig::new().with_data_transform(DataTransform::builder().pack());
-    let mut transfer = DmaTransfer::memory_to_memory(
-        config,
-        &mut channel,
-        src,
-        &mut dest_buffer,
-    );
+    let mut transfer =
+        DmaTransfer::memory_to_memory(config, &mut channel, src, dest);
 
     transfer.start().unwrap();
     transfer.wait_for_transfer_complete().unwrap();
+    let (_, dest) = transfer.free();
+
     let expected = [0x12345678; 10];
-    assert_eq!(expected, (*dest_buffer.data));
-    assert_eq!(expected, (*dest_buffer.data));
+    assert_eq!(expected, *dest);
+    assert_eq!(expected, *dest);
 
     log::info!("All tests passed!");
     loop {
--- a/src/gpdma.rs
+++ b/src/gpdma.rs
@@ -82,6 +82,7 @@ use crate::{
 };
 use core::{
     marker::PhantomData,
+    mem,
     ops::Deref,
     sync::atomic::{fence, Ordering},
 };
@@ -252,12 +253,8 @@ where
     D: WriteBuffer<Word: Word>,
 {
     channel: &'a mut CH,
-
-    // Hold onto source and destination for the lifetime of the transfer to ensure that any
-    // user implementations of ReadBuffer and WriteBuffer that are themselves references are held
-    // for the duration of the transfer and so prevent access to the underlying data.
-    _source: S,
-    _destination: D,
+    source: S,
+    destination: D,
 }
 
 impl<'a, CH, S, D> DmaTransfer<'a, CH, S, D>
@@ -289,8 +286,8 @@ where
 
         Self {
             channel,
-            _source: source,
-            _destination: destination,
+            source,
+            destination,
         }
     }
 
@@ -501,13 +498,12 @@ where
 
     /// Blocks waiting for a transfer to complete. Returns an error if one occurred during the
     /// transfer.
-    pub fn wait_for_transfer_complete(self) -> Result<(), Error> {
+    pub fn wait_for_transfer_complete(&mut self) -> Result<(), Error> {
         let result = self.channel.wait_for_transfer_complete();
         // Preserve the instruction and bus sequence of the preceding operation and
         // the subsequent buffer access.
         fence(Ordering::SeqCst);
 
-        core::mem::forget(self); // Prevents self from being dropped and attempting to abort
         result
     }
 
@@ -529,8 +525,28 @@ where
     }
 
     /// Abort a transaction and wait for it to suspend the transfer before resetting the channel
-    pub fn abort(self) {
-        // Allow Drop implementation to handle transfer abortion
+    pub fn abort(&mut self) {
+        if self.is_running() {
+            self.channel.abort();
+        }
+
+        self.disable_interrupts();
+
+        // Preserve the instruction and bus sequence of the preceding operation and
+        // the subsequent buffer access.
+        fence(Ordering::SeqCst);
+    }
+
+    pub fn free(mut self) -> (S, D) {
+        self.abort();
+        let (src, dest) = unsafe {
+            (
+                core::ptr::read(&self.source),
+                core::ptr::read(&self.destination),
+            )
+        };
+        mem::forget(self);
+        (src, dest)
     }
 }
 
@@ -541,14 +557,6 @@ where
     D: WriteBuffer<Word: Word>,
 {
     fn drop(&mut self) {
-        if self.is_running() {
-            self.channel.abort();
-        }
-
-        self.disable_interrupts();
-
-        // Preserve the instruction and bus sequence of the preceding operation and
-        // the subsequent buffer access.
-        fence(Ordering::SeqCst);
+        self.abort();
     }
 }


examples/dma.rs
```diff
@@ -0,0 +1,204 @@
...
+use cortex_m_rt::entry;
+use cortex_m_semihosting::debug;
+use stm32h5xx_hal::{
+    gpdma::{config::transform::*, DmaConfig, DmaTransfer},
+    pac,
+    prelude::*,
+};
+
+static mut SOURCE_BYTES: MaybeUninit<[u8; 40]> = MaybeUninit::uninit();
+static mut DEST_BYTES: MaybeUninit<[u8; 40]> = MaybeUninit::zeroed();
```
usbalbin, line 18: Why are these `MaybeUninit`? Also would `cortex_m::singleton!` work instead of these statics, getting rid of lots of the pointer gymnastics?

src/gpdma/config/transform.rs
```diff
@@ -0,0 +1,319 @@
...
+        self.transform.padding_alignment = PaddingAlignmentMode::Unpacked;
+        self
+    }
+}
+
+impl TruncationAlignment for DataTransformBuilder<u32, u8> {
+    /// Keep the least significant 8-bits and truncate the rest (default)
+    ///
+    /// ie: B7B6B5B4,B3B2B1B0 -> B4,B0
+    fn right_align_left_truncate(mut self) -> DataTransformBuilder<u32, u8> {
```
usbalbin, line 221: Nit, the other impl blocks has `Self` here
```suggestion
    fn right_align_left_truncate(mut self) -> Self {
```

examples/dma.rs
astapleton: This example is also old, when the DmaTransfer required `'static` buffers, which it no longer does, so I can actually reduce the complexity significantly here.

src/gpdma/ch.rs
```diff
@@ -0,0 +1,774 @@
...
+                .clear()
+                .tof()
+                .clear()
+        });
+
+        interrupt_clear_clock_sync_delay!(self.sr());
+    }
+
+    #[inline(always)]
+    fn check_transfer_event(
```
usbalbin, line 221: Do we want some comment saying what `Ok(true)` vs `Ok(false)` means?

src/gpdma/ch.rs
```diff
@@ -0,0 +1,774 @@
...
+    /// Aborts an operation by suspending the transfer and resetting the channel.
+    fn abort(&self);
+
+    /// Blocks waiting for a transfer to be started (or for it to be idle and complete). Reports any
+    /// errors that occur while waiting for the transfer to start.
+    fn wait_for_transfer_started(&self) -> Result<(), Error>;
+
+    /// Blocks waiting for a transfer to complete. Reports any errors that occur during a transfer.
+    fn wait_for_transfer_complete(&self) -> Result<(), Error>;
+    /// Blocks waiting for a half transfer event to trigger. Reports any errors that occur during a
```
usbalbin, line 489:
```suggestion

    /// Blocks waiting for a half transfer event to trigger. Reports any errors that occur during a
```

examples/dma.rs
```diff
@@ -0,0 +1,113 @@
...
+    log::info!("u32 to u32 with data transform");
+    let src = [0x12345678u32; 10];
+    let dest = &mut [0u32; 10];
+    let config = DmaConfig::new().with_data_transform(
+        DataTransform::builder()
+            .swap_destination_half_words()
+            .swap_destination_half_word_byte_order(),
+    );
+
+    let transfer = DmaTransfer::memory_to_memory(config, &channel, &src, dest);
```
usbalbin, line 52: Is this sound with things like Leakpocalypse in mind? As in, what if transfer is dropped/leaked after being started? Is there some precedence for things like this?

astapleton: What's Leakpocalypse? Dropping is problematic you're right. I'll add a drop implementation that aborts.

src/gpdma/config.rs
```diff
@@ -0,0 +1,412 @@
+use core::marker::PhantomData;
+
+use super::Word;
+
+pub mod transform;
+use transform::*;
+
+/// PeripheralRequests is used for peripheral-to-peripheral transfers to indicate which side of the
+/// transfer is driving the request (ie. which has the hardware request assigned)
+#[derive(Clone, Copy, Debug, PartialEq)]
```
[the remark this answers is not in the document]
astapleton: Thanks. I keep forgetting `defmt` derives!

examples/dma.rs
usbalbin: Ill try to find some source

usbalbin: But TLDR, `mem::leak` is safe so we can not rely on `Drop` for safety

usbalbin: Some info for a similar case https://doc.rust-lang.org/nomicon/leaking.html#threadscopedjoinguard

usbalbin: Some more specific info https://docs.rust-embedded.org/embedonomicon/dma.html . Not sure what for example embassy does though

astapleton: Ok. I see the problem. The drop implementation, and the improvements to wait_for_[half_]transfer_complete I added to consume the value (and avoid dropping) help in the case where `DmaTransfer` is actually dropped. It can't do anything in the case where someone calls `core::mem::forget` on the value. Doing so and returning from a function would trample over any memory which has now gone out of scope. 

On the one hand, this is pretty degenerate design. However, it is safe, and is very similar to the use case mentioned above. The obvious solution to avoid this use case (I think) is to require that the buffers have `'static` lifetimes, which I was trying to avoid because dealing with static variables is a real pain (although that `cortex_m::singleton` macro really helps!). This probably explains the `embedded-dma` library requiring static lifetimes for data.

Any ideas for avoiding the requirement to have `'static` lifetimes?

usbalbin: The scoped threads thing ended up like this https://blog.rust-lang.org/2022/08/11/Rust-1.63.0/#scoped-threads not sure if any of that is relevant here though...

[Here](https://github.com/embassy-rs/embassy/blob/e145a653cf9d31f101b0735406b8b7c9208bc1da/embassy-stm32/src/dma/dma_bdma.rs#L598-L600) is embassy's solution. From what I can tell from a quick glance (I might very well be missing something) they simply assume the value wont be leaked

astapleton: Yeah, I've been referencing the embassy implementation (hence not thinking about leaking!). This is a pretty good read: https://docs.rust-embedded.org/embedonomicon/dma.html. It mentions using `StableDeref`/`Pin` (which the embedded-dma traits actually require), which looks more forgiving than just `'static` lifetimes (prior to my previous understanding), so that's probably the way to go for the memory buffers.

astapleton: > (I have very limited experience with DMA and the weird safety issues it may cause. The above is just some info I have stumbled upon)

This is an excellent catch, so I wouldn't be too modest!

examples/dma.rs
astapleton: So I've reverted the change I made to this example after forcing immovable buffers. Unfortunately, `cortex_m::singleton!` gives me buffers with unaligned addresses, and I'm unsure how to address that. Open to ideas!

usbalbin: What? As in 

```rust
let foo: &mut [0u32; 10] = singleton!(FOO_BUFFER: [u32; 10] = [0u32; 10]);
assert!((foo as *const u32).is_aligned());
```

astapleton: I think I must have been tired yesterday. In order to get a reference to the buffer that I could check I was duplicating the reference with:
```
    let dest = singleton!(U8_DEST: [u8; 40] = [0u8; 40]).unwrap();
    let dest_copy = unsafe { *(dest.as_ptr() as *const &[u8]) };
```
which was giving me a pointer misalignment error. However, I realized this morning that that second line is incorrect because it's trying to acquire a fat pointer to a slice, which is incorrect.

Changing it this works fine:
```
    let dest_copy = unsafe { &*(dest.as_ptr() as *const [u8; 40]) };
```

So let me revise the example again to use the singleton pattern because it is much cleaner.

astapleton: Ok. Fixed it.

src/gpdma.rs
```diff
@@ -0,0 +1,550 @@
...
+//! - Memory to memory
+//! - Memory to peripheral
+//! - Peripheral to memory
+//! - Peripheral to peripheral
+//!
+//! Each GPDMA has 8 channels. Each channel can service any hardware request (or memory to memory
+//! transfer) that is supported by the processor (ie. they're not tied to specific channels). All
+//! channels support direct and linked-buffer transfers. However, the channels do have different
+//! capabilities (see RM0492 Rev 3 section 15.3.2 for full details), notably that channels 0-5 can
+//! only service transfers in a linear address space, while channels 6 & 7 can service transfers
```
usbalbin, line 15: Perhaps a clarification?

```suggestion
//! only service transfers in a linear address space, while channels 6 & 7 can also service transfers
```

src/gpdma.rs
```diff
@@ -0,0 +1,550 @@
...
+            dest_size
+        } else {
+            // When the source data width is less than or equal to the destination data width, we
+            // just need to ensure that the destination buffer is large enough to hold all of the
+            // source data.
+            assert!(src_size <= dest_size, "Transfer size ({src_size} bytes) will overflow the destination buffer ({dest_size} bytes)!");
+            src_size
+        };
+
+        // We also need to ensure that the destination
```
usbalbin, line 313: What? :)

src/gpdma.rs
```diff
@@ -0,0 +1,550 @@
...
+        )
+        .apply_hardware_request_config(config)
+        .apply_peripheral_source_config(config)
+    }
+
+    /// Create a new peripheral-to-peripheral transfer with source and destination peripherals
+    /// provided.
+    pub fn peripheral_to_peripheral<S, D, T>(
+        config: DmaConfig<PeripheralToPeripheral<T>, S::Word, D::Word>,
+        channel: &'a CH,
```
usbalbin, line 370: Is this really ok to have a shared reference? Wouldn't that allow multiple transfers to write to the same channels registers?

src/gpdma.rs
```diff
@@ -0,0 +1,550 @@
...
+
+    /// Blocks waiting for the half transfer complete event. Returns an error if one occurred during
+    /// the transfer.
+    pub fn wait_for_half_transfer_complete(self) -> Result<(), Error> {
+        let result = self.channel.wait_for_half_transfer_complete();
+        // Preserve the instruction and bus sequence of the preceding operation and
+        // the subsequent buffer access.
+        fence(Ordering::SeqCst);
+
+        core::mem::forget(self); // Prevents self from being dropped and attempting to abort
```
usbalbin, line 514: Sorry I have not gotten far enough to get a complete picture yet.

With that said, when running the dma in circular mode this function would prevent getting the next batch of values since it consumes self?

examples/dma.rs
```diff
@@ -0,0 +1,118 @@
...
+    let rcc = dp.RCC.constrain();
+    let ccdr = rcc.sys_ck(250.MHz()).freeze(pwrcfg, &dp.SBS);
+
+    let channels = dp.GPDMA1.channels(ccdr.peripheral.GPDMA1);
+
+    log::info!("u8 to u8");
+    let src =
+        singleton!(: [u8; 40] = core::array::from_fn(|i| i as u8)).unwrap();
+    let dest = singleton!(: [u8; 40] = [0u8; 40]).unwrap();
+    let source_copy = unsafe { &*(src.as_ptr() as *const [u8; 40]) };
```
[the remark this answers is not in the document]
astapleton: Yeah, the point is to verify the contents of the buffer after the DMA transfer is complete. Requiring a static reference to the buffer in the DmaTransfer API and only getting a reference from the `singleton` macro, instead of a value, means I can't access the `src` reference to the buffer after passing it to the `memory_to_memory` function (the reference gets moved). It's UB, but IMO it is safe because the reference is not used until the DMA transfer is stopped. Is that incorrect?

This does reflect the pain of using this API now. It's really tricky to use without some sort of unsafe usage, and is why I initially avoided requiring static references to buffers in the first place. The alternative I've seen (e.g. [stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/dma/mod.rs#L1120)) to this is to return the source & destination values (whether they be `'static` references or otherwise when the transfer completes (from a `free` method or when `wait_for_transfer_complete` returns), but I generally don't like that API because it now forces users to use options to store these references in user code, or use unsafe themselves. Maybe that reflects the challenges of using DMA in general? 

I've also been figuring out the peripheral API and using `Future`s based on top of this API, and those both become more painful with this change too. But, if you think that is a more sound API, I can try out the alternative above and see how that interacts with Future/peripheral usage. 

You can see the WIP Futures and SPI DMA in #65 and #66. Would love feedback if you have ideas for improving/settling on an API!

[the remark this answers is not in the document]
astapleton: I will also add that ensuring that the API is sound (by requiring `'static` source/destination), and putting the onus on the user to ensure that their buffer use is sound seems like the appropriate approach. 

That all said, this example also demonstrates how user can easily get around the requirement to have static buffers (by doing this exact operation), so perhaps the requirement for `'static` lifetimes is actually more painful than useful. The user can introduce UB by leaking `DmaTransfer` with the old API, or by copying references like I've done here. Dealer's choice.

Given that the requirement to use a `'static` lifetime *encouraged* me to use this hack (again, happy to learn how to do this in a safer way!), and leaking `DmaTransfer` seems like a much more deliberately degenerate thing to do, perhaps we could instead take the approach a similar approach embassy takes, and mark the transfer initialization functions as unsafe. We could then include a warning to not leak the `DmaTransfer` struct in the safety note.

src/gpdma.rs
astapleton: Oh yeah, this is a mistake. I just copied the same behvaior for `wait_for_transfer_complete`, which doesn't make sense. Will fix.

examples/dma.rs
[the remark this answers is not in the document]
usbalbin: @AdinAck have you had the opportunity to think about it any more since we spoke last in the G4 issue about DMA? Do you have any input here or just in general?

Also @astapleton, as to
> It's UB, but IMO it is safe because the reference is not used until the DMA transfer is stopped. Is that incorrect?

Please correct if I am wrong but I am pretty sure that strictly speaking it is instant UB the moment the second reference is created . Due to the UB the resulting binary is just something that may or may not happen to seem to do the right thing or some thing completely different which may or may not change depending on compiler version/options/alignment of stars/nasal demons etc. Yet I am not at all sure what we are allowed to do instead for that case... Perhaps a raw pointer that is only turned into a reference after the DMA is done?

I have tried(with not so great results) to read up on some recent discussion on the subject of DMA, volatile read/write etc. [Here](https://github.com/rust-lang/unsafe-code-guidelines/issues/321#issuecomment-2894697770) is some info. With, I think, this being most relevant ('2' being the DMA case):

> For (2), one has to put a suitable fence between the DMA accesses and the final MMIO access. This will, at least for now, require inline assembly: RISC-V, for instance, has separate fences for communicating with regular cached memory and with MMIO memory. The logical justification of the inline assembly block will be that it forks off a thread that asynchronously copies the DMA memory, modeling what the hardware does. So there's no new API required here from our side.

Then there is [this](https://github.com/rust-lang/rust/issues/144351) stating
> [...] asm block does not act as a compiler fence [...]

Then there is [this](https://github.com/rust-lang/unsafe-code-guidelines/issues/347#issuecomment-1183544719), stating
> [...] So, if your program contains no atomic accesses, but some atomic fences, those fences do nothing. [...]

And this [comment](https://github.com/rust-lang/unsafe-code-guidelines/issues/347#issuecomment-1179590431)
>> aliasing analysis with &mut can cause reordering of code because it knows there can't be any aliasing pointers.
>
> That is true with all fences, none of them allows you to violate the aliasing assumptions.
If you are using volatile because that memory is shared with some other party (e.g. via MMIO), then you must also inform the type system of that sharing, by using shared references or raw pointers.

I am getting more confused the more I read :/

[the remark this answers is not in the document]
astapleton: So it sounds like we're talking about a few things here:

1. UB in the example duplicating references
2. `'static` references to source/destination buffers
3. fences for ensuring that all accesses to the buffers before DMA starts and after DMA ends are done strictly before/after the DMA operation (ie. ensuring the correct ordering of instructions)
 
For 1. yes, I'm aware that this is UB. My assumption was that because there are fences (the same used by the  embassy-stm32 DMA implementation) before/after the transfer starts/finishes, that the memory accesses would be fine. However, I'm not really familiar with the lower level semantics of how rust treats references, so I may be quite wrong about this. Happy to be corrected.

2. The `'static` requirements on the buffers is what drove me to use the UB mentioned above. The only way I know to avoid it is for the `DmaTransfer` struct to hold onto the references for the duration of the transfer and to release them afterwards. I don't love that API choice, but I can give it a go and see how it impacts peripheral driver consumers? I think the alternative is to drop the `'static` lifetime requirement on user buffers (as I had earlier), mark the initialization functions as unsafe, and include a safety note that the consumer should not leak the transfer. Thoughts?

Onto 3: is the implication that the fences used here are not sufficient? I don't really follow the discussion about volatile accesses to the affected buffers though. We're not actually accessing the peripheral TX/RX data registers at all: the addresses are being supplied to the DMA which accesses them.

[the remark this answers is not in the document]
astapleton: According to https://doc.rust-lang.org/reference/behavior-considered-undefined.html#r-undefined.alias, the behavior in the example is pretty clearly considered unsound (specifically: taking a reference to a mutable reference that is considered live, which, with a `'static` lifetime is for the duration of the program). 

So I can take a look at fixing that in the example, if possible, or modifying the API accordingly. I will also have a go at figuring out how one would interact with this API in a real program to inform API changes.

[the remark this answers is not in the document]
astapleton: I've also been misunderstanding something: the reason I can't use these buffer references again, is because the API takes an implementation of ReadBuffer and WriteBuffer as values, so the buffer references are being moved, and obviously can't be accessed afterwards. It was just confusing because the values being moved were `&'static mut` references. So it is an API problem?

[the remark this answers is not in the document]
astapleton: Ok. I think I addressed things. It was actually simple to wrap the static references that cortex_m::singleton! returns such that the data buffers can be accessed after the transfer, and I did so in the example with what seem like sound implementations of ReadBuffer/WriteBuffer?

What I realized doing that is that the references to `Buffer` would not be held by the transfer until it completes, so I updated the `DmaTransfer` struct to hold onto the source/destination values until the `DmaTransfer` is dropped. For implemetations that are passed by value (ie. a static reference returned by singleton!) nothing changes, as the values were consumed anyway. But for implementations that are passed by reference, these references will now be held for the duration of the transfer.

I think this is now sound. @usbalbin WDYT?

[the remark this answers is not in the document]
astapleton: And... since I'm holding onto source/destination, I decided that I might as well add a `free` method and not consume self in `wait_for_transfer_complete` and `abort` to allow getting them back, so one can pass by value or by reference like in the example in 5c587f3. This makes the example a lot cleaner.

examples/dma.rs
```diff
@@ -7,12 +7,52 @@ mod utilities;
...
+use embedded_dma::{ReadBuffer, WriteBuffer};
 use stm32h5xx_hal::{
-    gpdma::{config::transform::*, DmaConfig, DmaTransfer},
+    gpdma::{config::transform::*, DmaConfig, DmaTransfer, Word},
     pac,
     prelude::*,
 };
 
+// Buffer is used to manage a reference to a static buffer returned by the cortex_m::singleton!
+// macro and which can be with the DmaTransfer API (which requires passing ReadBuffer and
```
usbalbin: [...] which can be *used* with[...]?

examples/dma.rs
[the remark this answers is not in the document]
usbalbin: @AdinAck wrote
> [...] BufferAccess would be a &mut ... if the endpoint was "source" and a **& ... if the endpoint was "destination"**. [...]

According to https://doc.rust-lang.org/reference/behavior-considered-undefined.html#r-undefined.alias
> The exact aliasing rules are not determined yet, but here is an outline of the general principles: **&T must point to memory that is not mutated while they are live(except for data inside an [UnsafeCell<U>](https://doc.rust-lang.org/core/cell/struct.UnsafeCell.html))**

so we can not store the reference to the dst buffer in the transfer(unless using UnsafeCell)?

Further docs:

> [...] and &mut T must point to memory that is not read or written by any pointer not derived from the reference and that no other reference points to while they are live.

Is that ok since, from what I gather the DMA is considered another thread that writes to the memory? However it does so using an address derived from the `&mut`..

---

@AdinAck wrote:
> Related to what you guys were talking about with compiler fencing, I think the most reasonable solution would be to provide the buffer reference to this closure by `read_volatile`ing the buffer, even though we could just reference it directly, because that forces the compiler to properly order the reads/writes. Otherwise, it would have no reason to make sure that "read status flag and make sure transfer is inactive" precedes "read src/dest buffer".

So then we should only ever access the buffers using volatile access?

@AdinAck wrote:
> For a oneshot transfer, the transfer endpoints would be deconstructed:
>
> let TransferStage::Complete(res) = src.release() else { ... };
> let TransferStage::Complete(buf) = dest.release() else { ... };

That would then affect `buf` here too? Then forcing us to put it into some wrapper to prevent direct access?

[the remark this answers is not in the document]
AdinAck: > so we can not store the reference to the dst buffer in the transfer(unless using UnsafeCell)?

Sorry I didn't explain myself well. My belief is the transfer should *hold* the buffer, not a reference to the buffer. The references I was talking about were the input to a closure that can be used to access the buffer while a transfer is not active.



> Is that ok since, from what I gather the DMA is considered another thread that writes to the memory? However it does so using an address derived from the `&mut`..

I don't think so. By derived I think they mean the pointer in Rust was derived from the reference, so the compiler can keep track of its provenance. In the case of DMA, in the eyes of the compiler it is an invisible external influence that has the ability to violate the invariances Rust establishes for the mutation of memory. Which is why I think we can express that by only using the buffer via volatile access. But I thought more about that and I'm not sure exactly how that would look yet.

> So then we should only ever access the buffers using volatile access?

That is what I was saying yes, but I'm not quite sure how to do that yet. But the buffer being considered volatile memory (transiently) makes sense to me.

> That would then affect buf here too? Then forcing us to put it into some wrapper to prevent direct access?

Yes I want the transfer to *hold* the buffer(s), so consuming the transfer after it is complete would produce the buffer.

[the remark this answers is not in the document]
astapleton: Is there anything actionable on this PR regarding your discussion, @AdinAck @usbalbin?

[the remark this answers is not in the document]
usbalbin: @AdinAck Just so we are talking about the same thing, "hold" as in we move the buffer itself(not a reference) into the transfer? In the case the buffer is a simple array, we would somehow have to pin it(or the entire transfer) to ensure its address does not change, I think? I am not too familiar with `Pin`, but I believe I read somewhere that since, for example, array impl `Unpin` that defeats what we are trying to do. We would then have to use something like [PhantomPinned](https://doc.rust-lang.org/std/marker/struct.PhantomPinned.html) to prevent that?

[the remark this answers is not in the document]
usbalbin: > Is there anything actionable on this PR regarding your discussion, @AdinAck @usbalbin?

Not so sure at this point. I feel this is a bit out of my depth mem-safety-semantics-(insert more words)-wise. I am struggling to find any other dma implementation that is more correct than this according to the links I found. Yet the hole, is the transfer allowed to hold a reference to the buffer while the buffer is being written to, feels a bit uncertain. But I do not have a single concrete suggestion for what to do differently :)

I guess we could merge this as is (perhaps behind some feature flag), play around with it some. I see you have some things in the pipe depending on this. Perhaps then at some point, if @AdinAck has done some work on his end, we can come together with our experiences for further improvements?

usbalbin:
BTW, you really have put a lot of work into this, thank you @astapleton ! :)

examples/dma.rs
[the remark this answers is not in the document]
usbalbin: Yes, and correct me here, but I believe they prevent the use of owned arrays(if that was what @AdinAck was refering to)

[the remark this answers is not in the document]
usbalbin: > [...] I suspect I can come up with an application specific structure using some fancy ZSTs [...]

Interesting! Let us know if you figure it out :)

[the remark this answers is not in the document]
AdinAck: > `core::mem::forget`

💀 Well, that's quite the thorn in my side huh. I'll keep thinking but I'm pretty sure you're right and that destroys any and all invariances I would want to establish with the DMA transfer type.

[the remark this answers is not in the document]
usbalbin: Yeah, really hope they find some solution to the leak thing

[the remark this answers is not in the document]
AdinAck: On a lighter note, [this](https://docs.rust-embedded.org/embedonomicon/dma.html#compiler-misoptimizations) section of the embedonomicon brought me some peace of mind. The usage of the compiler fence and memory barrier instructions (or lack thereof) is very clear. I was especially happy to read:

> in the current implementation, volatiles happen to work just like relaxed atomic operations. There's work going on to actually guarantee this behavior for future versions of Rust.

So I will feel comfortable using compiler fences (and maybe a `dsb` just to be safe) to ensure the ordering of DMA operations is enforced.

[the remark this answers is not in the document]
astapleton: > > `core::mem::forget`
> 
> 💀 Well, that's quite the thorn in my side huh. I'll keep thinking but I'm pretty sure you're right and that destroys any and all invariances I would want to establish with the DMA transfer type.

Yeah. I had the same realization :)

PR #65, opened 2025-08-02
DMA Futures

astapleton:
This implements futures for GPDMA transfers by adding a wrapper for `DmaTransfer`, `DmaTransferFuture`, implementing Future for `DmaTransferFuture` in the `gpdma::future` module, which is created when a `DmaTransfer` is awaited via the `IntoFuture` trait. Under the hood, the implementation uses an `AtomicWaker` from the [`futures-util`](https://docs.rs/futures-util/latest/futures_util/) crate and channel interrupts to drive the future. Because the wakers need to be defined statically in order to be accessible from the channel interrupts, and to prevent accidental colliding implementations of interrupt handlers for the DMA channels, the futures implementation is hidden behind the `gpdma-future` feature flag.

### DIFF
--- a/src/gpdma/ch.rs
+++ b/src/gpdma/ch.rs
@@ -15,7 +15,7 @@ use super::{
     DmaConfig, Error, Instance, Word,
 };
 
-trait ChannelRegs: Sealed {
+pub(super) trait ChannelRegs: Sealed {
     #[allow(unused)] // TODO: this will be used for linked-list transfers
     fn lbar(&self) -> &LBAR;
     fn fcr(&self) -> &FCR;
@@ -741,11 +741,19 @@ where
     }
 }
 
+#[cfg(feature = "gpdma-futures")]
+pub use super::future::DmaChannel;
+
 /// DmaChannel trait provides the API contract that all GPDMA channels exposed to the user
 /// implement.
+// Note: This trait is defined differently (in the `future` module) when the `gpdma-futures`
+// feature is enabled, to support async operations.
+#[cfg(not(feature = "gpdma-futures"))]
 #[allow(private_bounds)]
 pub trait DmaChannel: Channel {}
 
+#[cfg(not(feature = "gpdma-futures"))]
+#[allow(private_bounds)]
 impl<DMA, CH, const N: usize> DmaChannel for DmaChannelRef<DMA, CH, N>
 where
     DMA: Instance,
--- a/src/gpdma/ch.rs
+++ b/src/gpdma/ch.rs
@@ -131,7 +131,7 @@ where
     DMA: Instance,
     CH: ChannelRegs,
 {
-    pub(super) fn new() -> Self {
+    pub(super) unsafe fn new_unsafe() -> Self {
         DmaChannelRef {
             _dma: PhantomData,
             _ch: PhantomData,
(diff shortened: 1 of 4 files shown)


src/gpdma/future.rs
```diff
@@ -0,0 +1,255 @@
...
+use super::{
+    ch::{
+        Channel, ChannelRegs, DmaChannel0, DmaChannel1, DmaChannel2,
+        DmaChannel3, DmaChannel4, DmaChannel5, DmaChannel6, DmaChannel7,
+        DmaChannelRef,
+    },
+    DmaTransfer, Error, Instance, Word,
+};
+
+// This trait is defined to be bound by ChannelWaker when futures are enabled via the
```
usbalbin, line 42: Are those supposed to be `//` and not `///`?

src/gpdma/future.rs
```diff
@@ -0,0 +1,255 @@
...
+    CH: ChannelRegs,
+    Self: Deref<Target = CH>,
+{
+    #[inline(always)]
+    fn handle_interrupt() {
+        // This creates a DmaChannelRef instance for channel N, which is be a duplicate to the
+        // DmaChannelRef instance that is held as a mutable reference in the DmaTransfer struct
+        // while a transfer is in progress. However, it is only used to disable the transfer
+        // interrupts for the channel and wake up the task that is waiting for the transfer to
+        // complete, which can be done safely.
```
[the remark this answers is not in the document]
astapleton: Yeah, I was thinking the same while writing it, so I think it makes sense to mark new as unsafe

src/gpdma/future.rs
astapleton: I think it also needs a `///` block for public consumption. This was a note differentiating the definition with and without the `gpdma-future` feature flag

src/gpdma/future.rs
```diff
@@ -0,0 +1,258 @@
...
+        //
+        // When the DmaTransfer struct is dropped, interrupts are disabled for the channel,
+        // preventing this interrupt from being triggered. Interrupts are only enabled when the
+        // transfer is awaited (calling IntoFuture::into_future).
+        let mut ch = unsafe { Self::new_unsafe() };
+
+        // This is a single volatile write to the channel's interrupt status register to disable
+        // interrupts for the channel so this interrupt doesn't trigger again while the transfer
+        // struct is being polled.
+        ch.disable_transfer_interrupts();
```
usbalbin, line 170: I have never written any futures on embedded, very interesting to see :) So this is all the `$GPDMA _CH $CH` ISRs do? Disable interrupts and wake the corresponding task to let it resume?

Are there any interrupt status flags we need to clear or is that done by this line?

src/gpdma/ch.rs
```diff
@@ -741,11 +741,19 @@ where
...
 }
 
+#[cfg(feature = "gpdma-futures")]
+pub use super::future::DmaChannel;
+
 /// DmaChannel trait provides the API contract that all GPDMA channels exposed to the user
 /// implement.
+// Note: This trait is defined differently (in the `future` module) when the `gpdma-futures`
+// feature is enabled, to support async operations.
+#[cfg(not(feature = "gpdma-futures"))]
```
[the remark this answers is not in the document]
usbalbin: Oh sorry :)

usbalbin (approved):
Very cool stuff!

This looks good to me now as far as I can tell :)

PR #68, opened 2025-08-07
Don't track settings.json

astapleton:


usbalbin:
Sorry for the delay.

What is the motivation behind this change?

PR #75, opened 2026-01-08
serial: Add asynchronous serial driver for U(S)ARTs

astapleton:
This adds basic asynchronous serial functionality for the UART and USART peripherals on the STM32H5 family. The public API for using it is implemented using the embedded-io traits, which allows for both non-blocking and blocking operation using the trait methods.

Serial functionality for the LPUART is not implemented in this change, nor is synchronous behavior on the USART peripherals.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -64,6 +64,8 @@
 //! [io::WriteReady]: https://docs.rs/embedded-io/0.6.1/embedded_io/trait.WriteReady.html
 //! [embedded-io]: https://docs.rs/embedded-io/0.6.1/embedded_io/
 
+use core::fmt::Display;
+
 use embedded_io::{Error as IoError, ErrorKind as IoErrorKind};
 
 pub mod config;
@@ -86,6 +88,13 @@ pub enum Error {
     Parity,
 }
 
+impl Display for Error {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl core::error::Error for Error {}
 
 impl IoError for Error {
     fn kind(&self) -> IoErrorKind {
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -1,0 +1,127 @@
+//! Serial communication using USART & LPUART peripherals
+//!
+//! The serial module provides both asynchronous and synchronous functionality using the USART
+//! and LPUART* (only asynchronous) peripherals. It also exposes some, but not all, of the advanced
+//! functionality provided by the USART peripheral.
+//!
+//! It supports both blocking and non-blocking usage. Non-blocking usage is provided via the
+//! embedded-hal-nb crate's [`Read`][serial::Read] and [`Write`][serial::Write] traits, while
+//! blocking usage is abstracted as an IO stream and is exposed via embedded-io traits:
+//! [`Read`][io::Read], [`ReadReady`][io::ReadReady], [`Write`][io::Write], and
+//! [`WriteReady`][io::WriteReady] are implemented.
+//!
+//! 7, 8, and 9-bit word sizes are supported. For 7- and 8-bit operation, transactions are performed
+//! with u8 values. For 9-bit operation transactions are performed with u16 values. The correct word
+//! size must be indicated to the compiler during initialization.
+//!
+//! # Usage
+//!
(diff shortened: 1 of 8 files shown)


src/serial/usart/usart_def.rs
```diff
@@ -0,0 +1,415 @@
...
+                gpio::PB1<Alternate<7>>,
+                gpio::PC4<Alternate<7>>,
+                gpio::PD9<Alternate<7>>,
+                gpio::PC11<Alternate<7>>
+            ]
+            CK: [
+                NoCk,
+                gpio::PB12<Alternate<7>>,
+                gpio::PC12<Alternate<7>>,
+                gpio::PD10<Alternate<13>>
```
usbalbin: A can not see PD10 having any connection on AF13 on stm32h523, I have not checked the other MCUs

usbalbin: It does, however on AF7

src/serial/usart/usart_def.rs
```diff
@@ -0,0 +1,415 @@
...
+            ]
+        UART4:
+            TX: [
+                NoTx,
+                gpio::PA0<Alternate<8>>,
+                gpio::PA12<Alternate<6>>,
+                gpio::PB9<Alternate<8>>,
+                gpio::PC10<Alternate<8>>,
+                gpio::PD1<Alternate<7>>,
```
usbalbin: H523 has AF8 here for UART4_TX

src/serial/config.rs
```diff
@@ -0,0 +1,228 @@
...
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum StopBits {
+    #[doc = "1 stop bit"]
+    Stop1,
+    #[doc = "0.5 stop bits"]
+    Stop0p5,
+    #[doc = "2 stop bits"]
+    Stop2,
+    #[doc = "1.5 stop bits"]
+    Stop1p5,
```
usbalbin, line 40: What do you think about sorting these?

src/serial/usart.rs
```diff
@@ -0,0 +1,600 @@
...
+
+            // Shift USARTDIV[3:0] right by 1 when oversampling by 8
+            let frac = div & 0xF;
+            let div = (div & !0xF) | (frac >> 1);
+            (true, div)
+        } else {
+            return Err(config::InvalidConfig);
+        };
+
+        #[cfg(feature = "log")]
```
usbalbin: what about defmt?

src/serial.rs
```diff
@@ -0,0 +1,121 @@
...
+//! let bytes_written = serial.write(&[0x00, 0x11, 0x22]).unwrap();
+//! let buf = &mut [0u8; 3];
+//! let bytes_read = serial.read(buf).unwrap();
+//! ```
+//!
+//! * Note: Synchronous and LPUART functionality is not yet implemented
+//!
+//! # Examples
+//!
+//! - [Simple Blocking Example](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/examples/serial.rs)
```
usbalbin: Here is a link to the H7-hal, was that intended

src/serial.rs
```diff
@@ -0,0 +1,121 @@
...
+impl core::error::Error for Error {}
+
+impl IoError for Error {
+    fn kind(&self) -> IoErrorKind {
+        match self {
+            Error::Overrun => IoErrorKind::Other,
+            Error::Framing => IoErrorKind::Other,
+            Error::Noise => IoErrorKind::Other,
+            Error::Parity => IoErrorKind::Other,
+        }
```
usbalbin: If it is always `IoErrorKind::Other` then how about this?

```suggestion
        IoErrorKind::Other
```

src/serial/config.rs
astapleton: They're currently sorted in the order they're shown in the U(S)ART->CR2 register documentation. Would you prefer in numerical order?

src/serial/usart.rs
astapleton: We need a better way of managing this, but I've added in logging for defmt too

src/serial/config.rs
usbalbin: Nah, in that case I am fine with either

astapleton:
@usbalbin Any other concerns here?

usbalbin (approved):
Nice work!

PR #76, opened 2026-02-17
timer: Add basic timer operation for H5 family

astapleton:
This implements timer functionality for the H5 family. It implements functionality that is common to all timers on all MCUs in the family (although it does not implement any LPTIM peripherals). It builds on top of #59, but instead of using macros to define all timer methods, this instead defines traits that represent just the functionality exposed for basic timer operation (inspired by how the STM32F4 HAL is structured), and uses macros to implement those limited methods. This keeps the macro surface area minimized. Future development can define Traits for General and Advanced timers to add in functionality for those timer peripherals.

### DIFF
--- a/examples/blinky_timer.rs
+++ b/examples/blinky_timer.rs
@@ -1,0 +1,84 @@
+#![deny(warnings)]
+#![no_main]
+#![no_std]
+
+mod utilities;
+
+use core::{
+    cell::RefCell,
+    sync::atomic::{AtomicBool, Ordering},
+};
+
+use cortex_m::interrupt::Mutex;
+use cortex_m::peripheral::NVIC;
+use cortex_m_rt::entry;
+
+use stm32h5xx_hal::gpio::gpioa::PA5; // LED pin
+use stm32h5xx_hal::gpio::{Output, PushPull};
+use stm32h5xx_hal::{pac, pac::interrupt, prelude::*, timer};
+use utilities::logger::info;
+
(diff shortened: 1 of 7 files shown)


examples/blinky_timer.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    info!("Start blinking with timer...");
+    // Enable TIM2 interrupt
+    unsafe {
+        cp.NVIC.set_priority(interrupt::TIM2, 1);
+        NVIC::unmask::<interrupt>(interrupt::TIM2);
+    }
+
+    loop {
+        // do_nothing
```
usbalbin: Clippy does not like empty loops

src/timer/counters.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        // Start counter
+        self.tim.resume();
+    }
+
+    /// Configures the timer to count up at the given frequency
+    ///
+    /// Counts from 0 to the counter's maximum value, then repeats.
+    /// Because this only uses the timer prescaler, the frequency
+    /// is rounded to a multiple of the timer's kernel clock.
```
usbalbin: Perhaps relevant to document whether the rounding will be up down or closest?

astapleton: Good call. Done 👍

astapleton:
@usbalbin Would you mind taking another look at this? I'd love to merge it

usbalbin (approved):
Nice! LGTM

PR #77, opened 2026-04-09
Initialize I2C peripheral with custom timing value

leifanderson721:
This PR adds the ability to pass a custom value for the STM32's `TIMINGR` register when initializing an I2C peripheral. This adds the ability for the user to compute a value for the I2C timing calculations that is custom to their board. The idea is that this will provide more accurate values than the algorithm used when initializing the peripheral with only the desired bus frequency.

TIMINGR values can be generated using the official STM32CubeMX tool based on desired clock frequencies and measured rise and fall times. ST Micro also provides the AN4235 document and spreadsheet that uses a similar algorithm. Note that these algorithms use an iterative method to exhaustively search the space of possible values which is different from the one currently used in this HAL crate.

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -61,7 +61,8 @@
 //!
 //! let mut i2c_target = dp.I2C1.i2c_target_only(
 //!     (scl, sda),
-//!     TargetConfig::new(own_addr, bus_freq),
+//!     bus_freq,
+//!     TargetConfig::new(own_addr),
 //!     ccdr.peripheral.I2C1,
 //! );
 //! ```
@@ -125,7 +126,8 @@
 //!
 //! let mut i2c = dp.I2C1.i2c_controller_target(
 //!     (scl, sda),
-//!     TargetConfig::new(own_addr, bus_freq),
+//!     bus_freq,
+//!     TargetConfig::new(own_addr),
 //!     ccdr.peripheral.I2C1,
 //! );
 //! ```
@@ -318,13 +320,7 @@ pub struct ManualAck;
 /// Marker struct for I2cTarget implementation to indicate that it automatically handles all ACK'ing
 pub struct AutoAck;
 
-/// A raw pre-computed TIMINGR register value.
-///
-/// Wraps a `u32` to distinguish it from a bus frequency when passing to
-/// [`I2c::new`] or the [`I2cExt`] constructors. Obtain a value from STM32CubeMX
-/// or AN4235 and wrap it here to bypass runtime timing calculation.
-#[derive(Debug, Clone, Copy)]
-#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+/// A raw u32 representation of the TIMINGR register, which can be converted into an I2cTiming struct.
 pub struct RawTimingr(pub u32);
 
 /// Represents the TIMINGR register and its associated timing parameters for an I2C peripheral.
@@ -332,7 +328,7 @@ pub struct RawTimingr(pub u32);
 /// or calculated at runtime from the peripheral clock and target bus frequency.
 #[derive(Debug, Copy, Clone)]
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-pub struct TimingRegister {
+pub struct I2cTiming {
     presc: u8,
     scldel: u8,
     sdadel: u8,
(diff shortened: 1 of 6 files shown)


src/i2c.rs
```diff
@@ -605,74 +665,128 @@ fn calc_timing_params(ker_ck: u32, target_freq: u32) -> (u8, u8, u8, u8, u8) {
...
+///
+/// Types that encode all timing information directly (raw register values, explicit timing
+/// parameters, or a pre-built [`TimingRegister`]) ignore `ker_ck`. Types that specify only
+/// a desired bus frequency (e.g. [`Hertz`]) use `ker_ck` to calculate the timing parameters.
+pub trait IntoTimingRegister: Copy {
+    fn into_timing(self, ker_ck: u32) -> TimingRegister;
+}
+
+/// Raw TIMINGR register value — parsed directly into timing fields.
+impl IntoTimingRegister for RawTimingr {
```
astapleton: Similar to my comment about allowing a u32 to be provided instead of a Hertz object for specifying timing by bitrate, I don't really love the idea of just wrapping a raw u32 bitfield like this and passing it into the I2C creation API. 

Is this what CubeMX spits out for the TIMINGR value? If so, I think it would be more explicit/intentional if, instead of allowing this to be passed into the I2C API directly, it was first explicitly parsed into a TimingRegister struct via a From implementation, or via the new method (explicitly assigning each field):

```
let timing = TimingRegister::from(u32);
let mut i2c =
        dp.I2C2
            .i2c((scl, sda), timing, ccdr.peripheral.I2C2, &ccdr.clocks);
```

src/i2c.rs
```diff
@@ -605,74 +665,128 @@ fn calc_timing_params(ker_ck: u32, target_freq: u32) -> (u8, u8, u8, u8, u8) {
...
 
-    if let Some(secondary_address) = config.secondary_address {
-        i2c.oar2().write(|w| {
-            w.oa2().set(secondary_address);
-            if let Some(mask) = config.secondary_address_mask_bits {
-                w.oa2msk().set(mask + 1); // The address is shifted up by one, so increment the mask bits too
-            }
-            w
-        });
+impl IntoTimingRegister for u32 {
```
astapleton: I prefer requiring that the caller specifically provide a Hertz type in the case of specifying the desired bus rate instead of TimingRegister type. Allowing for a u32 when it's possible to use a RawTimingr is a little ambiguous/confusing (I was confused initially at what this was for).

src/i2c.rs
```diff
@@ -318,6 +318,28 @@ pub struct ManualAck;
...
+#[derive(Debug, Clone, Copy)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct RawTimingr(pub u32);
+
+/// Represents the TIMINGR register and its associated timing parameters for an I2C peripheral.
+/// This can be instantiated from a constant calculated elsewhere (CubeMX or AN4235)
+/// or calculated at runtime from the peripheral clock and target bus frequency.
+#[derive(Debug, Copy, Clone)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct TimingRegister {
```
astapleton: I don't like the naming of this. It a leaky abstraction. I think I'd rather name it something like `I2cTiming` as other HALs have done, or something similar, that doesn't expose that it maps directly to a specific registers in the current hardware implementation. These peripherals are frequently updated in newer versions of these chips, and later versions might specify an additional register for timing, which we could also encompass in this register. A more substantial update might split these parameters across multiple registers (or do away with them entirely).

astapleton: Can you create a new function for this so that it's possible to explicitly create one from calling code?

src/i2c.rs
```diff
@@ -605,74 +665,128 @@ fn calc_timing_params(ker_ck: u32, target_freq: u32) -> (u8, u8, u8, u8, u8) {
...
+            sdadel: ((timingr >> 16) & 0x0F) as u8,
+            sclh: ((timingr >> 8) & 0xFF) as u8,
+            scll: (timingr & 0xFF) as u8,
         }
-        AddressMode::AddressMode10bit => {
-            w.oa1().set(config.own_address).oa1mode().bit10()
+    }
+}
+
+impl IntoTimingRegister for (u8, u8, u8, u8, u8) {
```
astapleton: This isn't currently supported in the API, and I'd rather the TimingRegister struct is created explcitly with a new function encapsulating these values, instead of implicitly in the API call to create a new I2c instance.

src/i2c.rs
```diff
@@ -605,74 +665,128 @@ fn calc_timing_params(ker_ck: u32, target_freq: u32) -> (u8, u8, u8, u8, u8) {
...
     /// # Panics
     ///
-    /// Panics if the ratio between `frequency` and the i2c_ker_ck
+    /// Panics if the ratio between the bus frequency and the i2c_ker_ck
     /// is out of bounds. The acceptable range is [4, 8192].
     ///
-    /// Panics if the `frequency` is too fast. The maximum is 1MHz.
-    pub fn new(
+    /// Panics if the bus frequency is too fast. The maximum is 1MHz.
+    pub fn new<Timing: IntoTimingRegister, CfgTiming: IntoTimingRegister>(
```
astapleton: The convention for generic parameters is to use CAPs. This is how it's used elsewhere in this HAL too.

src/i2c.rs
```diff
@@ -1573,8 +1689,11 @@ impl<I2C: Instance> Targetable for I2c<I2C, SwitchRole> {
         self.inner.disable_target_event(event)
     }
 
-    fn configure_target(&mut self, config: TargetConfig) {
-        configure_target_addresses(&self.i2c, config)
+    fn configure_target<T: IntoTimingRegister>(
```
astapleton: this is an internal method. It doesn't need to take a generic parameter. The `into_timing` method can be called before it is passed into this function

src/i2c/config.rs
```diff
@@ -18,23 +18,26 @@ pub struct TargetConfig {
     // Address mask for secondary mask
     pub(crate) secondary_address_mask_bits: Option<u8>,
     /// Frequency at which bus is expected to run.
-    pub(crate) bus_frequency_hz: u32,
+    pub(crate) timing: T,
```
astapleton: This struct being generic has a lot of knock-on effects, evident throughout this PR. 

If we were to store this timing information in the config struct, I think it would be better to store the TimingRegister struct directly. However, given that we can't derive a TimingRegister from a bus frequency without a kernel clock value, I think it would actually be better to remove the bus frequency from the `TargetConfig` struct entirely, and require that it is passed into the `I2c` creation method on it's own, reducing footprint of this generic requirement significantly. 

I acknowledge that it is a breaking API change, but I think we can afford that at this point, given that we haven't published a crate yet!

src/i2c.rs
```diff
@@ -367,122 +389,160 @@ impl<I2C> Inner<I2C> {
...
         clocks: &CoreClocks,
     ) -> I2cTarget<I2C, AutoAck, SingleRole>;
 
     /// Create an I2c instance that can switch roles to a I2cTarget to perform
     /// Target operations
-    fn i2c_controller_target<P: Pins<I2C>>(
+    fn i2c_controller_target<
+        P: Pins<I2C>,
+        T: IntoTimingRegister,
+        CT: IntoTimingRegister,
```
astapleton: Given https://github.com/stm32-rs/stm32h5xx-hal/pull/77/changes#r3065875784, this is moot, but I wanted to note that it isn't a good idea to allow for 2 different types to be specified for timing and the IntoTimingRegister implementation in TargetConfig. 

Noticing this further reinforces my opinion in https://github.com/stm32-rs/stm32h5xx-hal/pull/77/changes#r3065875784 that we remove timing information from the TargetConfig struct. It shouldn't have been there in the first place as it allows (in this API in its current form) for 2 bus frequencies to be specified for the bus. And the one stored in TargetConfig isn't even used when switching role from Target to Controller, which is pretty confusing if you were intending that to be so by specifying it in the TargetConfig. Removing it from TargetConfig improves the clarity of the API and doesn't give the impression that the driver does something that it doesn't.

astapleton (changes requested):
I have a few concerns about this currently, some stemming from poor API choices I made in the past. See comments

src/i2c.rs
```diff
@@ -1737,9 +1609,84 @@ impl<I2C> I2cTarget<I2C, SwitchRole> {
...
+}
+
+/// A type that can be converted into [`I2cTiming`].
+///
+/// Implemented for:
+/// - [`I2cTiming`] — used as-is
+/// - [`RawTimingr`] — decoded from a raw TIMINGR register value (e.g. from STM32CubeMX)
+/// - `(Hertz, Hertz)` — `(ker_ck, bus_freq)` tuple; computes timing at runtime
+pub trait IntoI2cTiming {
+    fn into_i2c_timing(self) -> I2cTiming;
```
astapleton: why did you change the signature of this from the previous commit to no longer take the kernel clock? This forces the change in the API for creating an I2c instance from specifying a Hertz value to specifying a tuple of Hertz and kernel clock.

src/i2c.rs
```diff
@@ -605,74 +636,41 @@ fn calc_timing_params(ker_ck: u32, target_freq: u32) -> (u8, u8, u8, u8, u8) {
...
                 .sdadel()
-                .set(sdadel)
+                .set(timing.sdadel)
                 .scldel()
-                .set(scldel)
+                .set(timing.scldel)
         });
 
-        if let Some(config) = config {
+        if let Some(ref config) = config {
```
astapleton: what's this change for?

src/i2c.rs
```diff
@@ -1615,9 +1609,84 @@ impl<I2C> I2cTarget<I2C, SwitchRole> {
     }
 }
 
+/// Functions for calculating timing parameters for the I2C peripheral
+impl I2cTiming {
```
astapleton, line 1656: Can you group these impl blocks together with the I2cTiming struct?

src/i2c.rs
```diff
@@ -1615,9 +1609,84 @@ impl<I2C> I2cTarget<I2C, SwitchRole> {
...
+        I2cTiming {
+            presc,
+            scldel,
+            sdadel,
+            sclh,
+            scll,
+        }
+    }
+
+    fn calc_register_value(&self) -> u32 {
```
astapleton: return RawTimingr? perhaps implement this as a `From<I2cTiming> for RawTimingr`?

src/i2c.rs
```diff
@@ -367,122 +385,135 @@ impl<I2C> Inner<I2C> {
...
         target_config: TargetConfig,
         rec: I2C::Rec,
-        clocks: &CoreClocks,
     ) -> I2c<I2C, SwitchRole> {
-        I2c::new(self, frequency, Some(target_config), rec, clocks)
+        I2c::new(self, timing.into_i2c_timing(), Some(target_config), rec)
+    }
+}
+
+fn configure_target_addresses<I2C: Instance>(i2c: &I2C, config: &TargetConfig) {
```
astapleton: nit: avoid moving things around if not necessary. Moving this function just bloats the diff

astapleton (changes requested):
This looks better, but ideally the API would facilitate that examples don't change. The requirement to pass in the kernel clock to provide a bus frequency as the parameter seems unnecessary.

src/i2c.rs
leifanderson721: This seems like a clearer way to specify the arguments required to perform the runtime calculation of the timing values and echoing the pattern used to pass the pin values in. This also makes it clearer in the internal code which values are being used to generate the timing. 

I think the examples will need to change if we want to move the timing parameter out of the `TargetConfig` struct. My commit before this was an attempt to match the examples exactly but your earlier comment said to change the API.

src/i2c.rs
leifanderson721: This is left over from an earlier version. where `TargetConfig` took a generic timing parameter.

src/i2c.rs
astapleton: The idea of preserving the ability to specify simply the bus frequency is to keep the simple API simple. It's also unnecessary to pass in the kernel clock because reading the kernel clock is already taken care of in the driver. Specifying the I2cTiming struct is a more advanced use case, so the added explicitness that I've asked for is to ensure that the caller knows that and knows what they're doing.

> I think the examples will need to change if we want to move the timing parameter out of the TargetConfig struct. My commit before this was an attempt to match the examples exactly but your earlier comment said to change the API.

Ah, you're right that the target examples will need to change to take the extra timing parameter. However, they can remain the same otherwise without the need to extract the kernel clock manually, if you keep the clocks as a parameter, and specify the kernel clock in the IntoI2cTiming trait method.

astapleton (approved):

PR #78, opened 2026-04-22
crc: Add hardware CRC unit

leifanderson721:
The hardware CRC offers the ability to feed in data and have the hardware compute a rolling checksum used for error-detection.

This takes liberally from the implementation in
[stm32h7xx-hal](https://github.com/stm32-rs/stm32h7xx-hal).

### DIFF
--- a/src/crc.rs
+++ b/src/crc.rs
@@ -260,7 +260,7 @@ impl ReverseInput {
 
 /// A CRC polynomial.
 ///
-/// The STM32H7 CRC unit only supports odd polynomials, and the constructors
+/// The STM32 CRC unit only supports odd polynomials, and the constructors
 /// will check to ensure the polynomial is odd unless the `_unchecked` variants
 /// are used.
 ///
--- a/src/crc.rs
+++ b/src/crc.rs
@@ -137,18 +137,6 @@ impl Crc {
         result
     }
 
-    /// Write to the independent data register
-    /// This is "temporary storage location for four bytes" and is not affected by reset
-    /// It is also not used in the CRC calculation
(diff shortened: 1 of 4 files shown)


src/crc.rs
```diff
@@ -0,0 +1,403 @@
...
+            Self::ByteReverse => REV_IN::Byte,
+            Self::HalfWordReverse => REV_IN::HalfWord,
+            Self::WordReverse => REV_IN::Word,
+        }
+    }
+}
+
+/// A CRC polynomial.
+///
+/// The STM32H7 CRC unit only supports odd polynomials, and the constructors
```
astapleton: Please scrub/replace references to STM32H7 (and its reference manual, if any)

src/crc.rs
```diff
@@ -0,0 +1,403 @@
...
+    /// This is "temporary storage location for four bytes" and is not affected by reset
+    /// It is also not used in the CRC calculation
+    pub fn set_idr(&mut self, value: u32) {
+        self.reg.idr().write(|w| w.idr().set(value));
+    }
+
+    /// Read from the independent data register
+    pub fn get_idr(&self) -> u32 {
+        self.reg.idr().read().idr().bits()
+    }
```
astapleton: Do we know what this is for? I looked at the original H7 PR and they didn't know. Seems unnecessary to expose it?

leifanderson721: According to the reference manual it's just a scratch register; "temporary storage location for four bytes".

astapleton: I don't know that it's really worth exposing in the API?

astapleton (approved):

